PR #3, opened 2026-04-08
feat: add SetTextElements to DocuClient interface

hf-mrdachner:
## Summary

- Add `SetTextElements` method to `DocuClient` interface for writing text symbols and selection texts via ADT PUT
- Remove discovery pre-check from `GetTextElements` — S4 supports the endpoint but doesn't advertise it in `/sap/bc/adt/discovery`, causing false negatives
- Add `formatTextSymbols` and `formatSelectionTexts` for building PUT request bodies
- Add roundtrip tests (format → parse) for both symbols and selections
- Update custexport mock to satisfy the expanded interface

Needed by Hochfrequenz/mcp-server-abap#243

## Test plan

- [x] `go test ./...` all passing
- [x] `go vet ./...` clean
- [x] Roundtrip tests verify format/parse consistency


hf-mrdachner:
## Integration test result + impl fixes — PR #3

The integration test exposed two real bugs in the original `SetTextElements` implementation, both fixed in commit \`e752044\`. A new integration test (commit \`0b88fc6\`) exercises the transport-required path.

### Implementation fixes (commit \`e752044\`)

The S/4 textelements PUT endpoint requires more than the original implementation sent:

- **\`Accept\` header matching \`Content-Type\`** — without it S/4 returns 400 \`ExceptionResourceBadRequest\`: "Accept header missing".
- **\`X-sap-adt-sessiontype: stateful\`** — same pattern as \`SetSource\`; keeps the request bound to the SAP work process holding the lock.

### Lock handling pattern (test-side discovery)

The textelements endpoint binds to a *different* SAP enqueue resource (\`REPT\`) than the program-level lock acquired via \`LockObject(programURI)\` (which gives \`ESRDIRE\`). Result: \`SetTextElements\` with the program lock returns 423 \`ExceptionResourceInvalidLockHandle\` — "REPT not locked".

Fix: lock the **textelements URI** directly (\`/sap/bc/adt/textelements/programs/<NAME>\`) before calling \`SetTextElements\`. Both new integration tests use this pattern.

### Integration test results

\`SAP_INTEGRATION_SYSTEMS=hfq,s4u\`

#### \`TestSetTextElements_MultiSystem\` ($TMP path, no transport, regular \`integration\` tag)

| System | Result | Detail |
|---|---|---|
| **R/3** (\`hfq\`) | SKIP | textelements endpoint not exposed on R/3 ECC for this object kind (404) |
| **S/4** (\`s4u\`) | ✅ PASS | read-back: 1 symbol, 1 selection from fresh \$TMP program |

#### \`TestSetTextElements_WithTransport_MultiSystem\` (transport-required path, \`integration && transport\` tag)

| System | Result | Detail |
|---|---|---|
| **R/3** (\`hfq\`) | SKIP | textelements endpoint not exposed |
| **S/4** (\`s4u\`) | ✅ PASS | created transport, created Z_TXTR_* in Z_ADT_MCP_TEST package, wrote with \`?corrNr=\`, verified, cleanup released transport |

### Cleanup

- \$TMP test deletes its own program on cleanup.
- Transport test releases its transport and deletes its disposable program on cleanup.
- No leftover objects or transports observed.

hf-mrdachner:
## Independent Review — PR #3

*Reviewer: Claude Sonnet 4.6 via superpowers:code-reviewer (no prior conversation context). Reviewed from a detached-HEAD worktree at the tip of \`origin/feat/set-text-elements\` (commit \`0b88fc6\`).*

### Build & lint results

All checks passed cleanly in the worktree:

- \`go test ./...\` — PASS (all packages)
- \`go build -tags integration ./adt/...\` — clean
- \`go build -tags 'integration transport' ./adt/...\` — clean
- \`go vet -tags integration ./adt/...\` — clean
- \`golangci-lint run --enable dupl,goconst,gocyclo ./...\` — 0 issues

### Strengths

- **Clean separation of concerns.** \`writeTextElementSource\` is a focused helper; \`SetTextElements\` composes it naturally, mirroring how \`GetTextElements\` composes \`readTextElementSource\`.
- **Headers are correct.** \`X-sap-adt-sessiontype: stateful\` is required to keep the lock alive across the PUT; \`Accept: <content-type>\` is the S/4 strictness fix discovered during integration testing. Both are well-motivated and well-commented.
- **Lock-URI pattern is documented.** The comment on \`writeTextElementSource\` explicitly states that the textelements endpoint uses \`?lockHandle=\` (not the header that \`SetSource\` tries first), which is the non-obvious correctness pattern discovered against S/4. This is exactly the kind of discovery that should be written down.
- **Discovery pre-check removal is justified.** S/4 supports the textelements endpoint but does not advertise it in \`/sap/bc/adt/discovery\`, so the old \`hasEndpointInDiscovery\` guard produced false negatives. The new code lets the HTTP response itself be the authority, which is the correct approach. The PR description explains this clearly.
- **Interface extension is complete.** \`DocuClient\`, \`ClientRegistry\`, and the \`custexport\` mock are all updated. No other \`DocuClient\` implementers exist in-repo.
- **Integration tests are well-structured.** Both tests follow the established \`eachSystem(t)\` pattern, include \`t.Cleanup\` for object and transport teardown, and skip gracefully on \`isEndpointUnavailable\`. The transport-required test is correctly isolated behind \`integration && transport\` build tags.
- **Roundtrip tests pass.** \`TestRoundtripTextSymbols\` and \`TestRoundtripSelectionTexts\` verify the format/parse inverse property.

### Findings

**Important: URL parameters are not percent-encoded in \`writeTextElementSource\`.**

\`writeTextElementSource\` builds the query string by direct string concatenation:

```go
path += sep + "lockHandle=" + lockHandle
// ...
path += sep + "corrNr=" + transport
```

The analogous \`setSourceWithLockParam\` in \`source.go\` uses \`url.Values\` with \`params.Encode()\`, which applies \`url.QueryEscape\`. SAP lock handles are opaque tokens that can contain characters that are special in URL query strings (e.g., \`+\` signs are documented in some ADT implementations). Transport numbers (\`NPLK000001\`) are safe in practice, but the lock handle format is not guaranteed. The inconsistency with the pattern established in \`source.go\` is a latent correctness risk.

The same issue exists in \`lock.go\`'s \`UnlockObject\` (pre-existing, not introduced here), but \`writeTextElementSource\` should follow the \`setSourceWithLockParam\` pattern and use \`url.Values\`.

**Minor: No httptest mock unit test for \`writeTextElementSource\` / \`SetTextElements\`.**

Every other mutating method in the codebase has at least one \`httptest.NewServer\`-based unit test that verifies the outgoing request shape (correct path, headers, method, body). \`SetTextElements\` has only format/parse tests and integration tests. This means the \`Accept\` header, \`X-sap-adt-sessiontype\` header, and \`?lockHandle=\` / \`?corrNr=\` URL parameters are not exercised by unit tests — a future refactor that dropped a header would only be caught by the integration suite. See \`adt/messageclass_http_test.go\` and \`adt/source_test.go\` for the expected pattern.

**Minor: \`TestRegistryDelegatesAllMethods\` does not cover \`SetTextElements\`.**

The \`allEndpointsHandler\` in \`registry_test.go\` returns 200 for unmatched paths, so \`SetTextElements\` would not panic, but the delegation is not explicitly exercised in that test. Other methods added to the \`DocuClient\` interface (e.g. \`SetMessages\`) similarly lack registry delegation tests — this is a pre-existing gap, not a regression introduced by this PR.

**Minor: Acceptance criterion \`Returns ErrTextElementsNotSupported…\` is not met.**

Issue #45 lists: *"Returns \`ErrTextElementsNotSupported\` when the discovery document does not advertise the collection endpoint, matching \`GetTextElements\`' behaviour."* The PR removes that check from \`GetTextElements\` and never adds it to \`SetTextElements\`. The PR description explains the rationale (S/4 doesn't advertise the endpoint even though it works), and the integration tests validate that the write path works correctly without the guard. The decision is defensible, but it diverges from the stated acceptance criterion without an explicit retraction in the PR. A note in the PR description explaining why this criterion was deliberately dropped would be appropriate.

### Summary

The core implementation — \`SetTextElements\`, \`writeTextElementSource\`, \`formatTextSymbols\`, \`formatSelectionTexts\` — is correct and reflects real-world behaviour discovered against both R/3 and S/4. The lock-URI pattern (locking the textelements URI, not the program URI) is a genuine correctness requirement that is now documented in code. The lint check passes cleanly.

The important finding (missing URL encoding for \`lockHandle\`) is a latent bug that works in practice because SAP lock handles appear to be alphanumeric-safe, but it is inconsistent with the pattern established elsewhere in the codebase and should be fixed to eliminate future fragility.

### Verdict: GO with notes

Merge is appropriate. Before or shortly after merge, consider:
1. **(Should fix)** Switch \`writeTextElementSource\` to use \`url.Values\` / \`params.Encode()\` for \`lockHandle\` and \`corrNr\`, consistent with \`setSourceWithLockParam\` in \`source.go\`.
2. **(Nice to have)** Add an \`httptest\`-based unit test for \`SetTextElements\` that verifies the PUT path, headers (\`Content-Type\`, \`Accept\`, \`X-sap-adt-sessiontype\`), and \`?lockHandle=\` / \`?corrNr=\` query parameters.
3. **(Nice to have)** Add a note to the PR description or a follow-up issue acknowledging the deliberate removal of the \`ErrTextElementsNotSupported\` guard from both \`GetTextElements\` and the new \`SetTextElements\`.

hf-mrdachner:
## Review notes addressed (commit \`1d0b5c5\`)

Following the independent reviewer's "GO with notes" verdict, all three actionable findings have been addressed:

**1. Important — URL encoding** ✅
\`writeTextElementSource\` now builds the query string via \`url.Values\` / \`params.Encode()\` instead of raw concatenation, matching the pattern in \`setSourceWithLockParam\` (source.go). The new \`TestSetTextElements_PercentEncodesURLParameters\` unit test guards the regression risk: it asserts that lock handle and transport values containing \`+\`, \`/\`, \`=\`, \`&\`, and spaces are properly percent-encoded.

**2. Minor — httptest mock unit test** ✅
Added \`adt/textelements_http_test.go\` with five httptest-based tests that verify the outgoing request shape:
- Both endpoints (symbols + selections) are PUT with correct vendor MIME types in \`Content-Type\` and \`Accept\`.
- \`X-sap-adt-sessiontype: stateful\` is sent on every PUT.
- \`?lockHandle=\` and \`?corrNr=\` are correctly placed in the query string.
- Symbols-only and selections-only calls emit only the corresponding PUT.
- Empty transport correctly omits \`corrNr\` from the URL.

**3. Minor — \`ErrTextElementsNotSupported\` acceptance criterion** (deliberately dropped)
The original issue #45 listed: *"Returns \`ErrTextElementsNotSupported\` when the discovery document does not advertise the collection endpoint."* This criterion is intentionally not met because:
- S/4 supports the textelements endpoint but does not advertise it in \`/sap/bc/adt/discovery\`, so a discovery-based guard produces false negatives (the symptom this PR's discovery pre-check removal in \`GetTextElements\` already addresses).
- The HTTP response itself is the authoritative signal: \`SetTextElements\` returns the SAP error transparently (\`SAP ADT error 404 (ExceptionResourceNotFound): ...\` on R/3 ECC), and callers can already distinguish "not supported" via \`*adt.ADTError\` + status-code checks.
- Adding a new sentinel error solely to satisfy the criterion would be a YAGNI surface — no in-repo caller has asked for it, and the error path is already typed and inspectable.

Worth tracking as a follow-up if downstream consumers (mcp-server-abap, etc.) end up wanting a sentinel.

## Verification

- \`go test ./...\` — green
- \`golangci-lint --enable dupl,goconst,gocyclo\` — 0 issues
- Integration test on s4u (\$TMP path) — still PASS

Issue #15, opened 2026-04-08
TABL ETag mismatch on S/4: client If-Match drops "; charset=utf-8" from server-supplied ETag


hf-kklein:
## Summary

\`SetSource\` for a freshly created TABL on S/4 fails with **HTTP 412 \`ExceptionPreconditionFailed\`**. The server's error message echoes both ETags verbatim and the difference is exactly **\`; charset=utf-8\`** embedded inside the ETag value:

| Source | ETag value (literal) |
|---|---|
| Client \`If-Match\` | \`20260408224017000text/plain2\` |
| Server | \`20260408224017000text/plain; charset=utf-8 2\` |

The leading timestamp \`20260408224017000\` and trailing version \`2\` match. SAP's ETag format for TABL apparently includes the response Content-Type with charset as part of the ETag value. Adtler is somewhere stripping the \`; charset=utf-8\` chunk before sending it back as \`If-Match\`.

## Reproduction (S/4, 1/1)

\`\`\`go
ctx := context.Background()
client := adt.NewClient(<S/4 cfg>)

err := client.CreateObject(ctx, "TABL", "ZTAB_MCPTEST_S1", "\$TMP", "test", "")
// → success

// Whichever path adtler uses internally to read the ETag and then PUT with If-Match
// dies with 412 because the stored / sent value is missing "; charset=utf-8":
\`\`\`

> SAP ADT error 412: ExceptionPreconditionFailed
> "Client ETag \`20260408224017000text/plain2\` does not match the object ETag \`20260408224017000text/plain; charset=utf-8 2\` in the server"
> T100KEY: \`SADT_RESOURCE 043\`
> com.sap.adt.communicationFramework.subType: \`IfMatch\`

## Where to look

The leading "20260408224017000" + trailing "2" survive intact, so the ETag isn't being recomputed from scratch — adtler is reading **something** and dropping just the \`; charset=utf-8\` chunk. Likely candidates:

- A \`mime.ParseMediaType\` / \`strings.SplitN(header, ";", 2)\` call that's used to extract the bare media type from a Content-Type and (incorrectly) re-injected into the ETag, instead of preserving the response \`ETag\` header verbatim.
- A normalization helper that quotes/unquotes ETags and accidentally truncates at the first \`;\`.
- Direct construction of the ETag from \`resp.Header.Get("Content-Type")\` in some TABL-specific code path (if adtler synthesizes ETags for DDIC tables instead of reading them from the server response).

[...]

Issue #4, opened 2026-04-08
`CreateObject` leaks `ESRDIRE` enqueue on S/4 — next call fails with 423 InvalidLockHandle


hf-kklein:
## Summary

`adt.Client.CreateObject` succeeds on S/4 but leaves a session-bound exclusive enqueue on `TRDIR/<name>` that the SAP server then refuses to release for any subsequent call from the same logical client. The next `LockObject` returns `403 ExceptionResourceNoAccess` and the next `SetSource` (with auto-lock) returns `423 ExceptionResourceInvalidLockHandle`. The lock survives indefinitely until the SAP work-process session itself dies (or the row is force-deleted via SM12). Confirmed not to decay within 5 minutes of waiting.

The bug is **S/4-specific.** On ECC the same call sequence works cleanly with zero orphan locks.

Downstream consumer report with full reproduction, SM12 evidence, screenshots and the side-by-side ECC/S/4 contrast: [Hochfrequenz/mcp-server-abap#282](https://github.com/Hochfrequenz/mcp-server-abap/issues/282).

## Reproduction

Against any S/4 on-prem system:

\`\`\`go
cfg := /* your S/4 config */
client := adt.NewClient(cfg)
ctx := context.Background()

err := client.CreateObject(ctx, "PROG", "Z_REPRO", "\$TMP", "repro", "")
// → success

_, err = client.LockObject(ctx, "/sap/bc/adt/programs/programs/z_repro")
// → 403 com.sap.adt/ExceptionResourceNoAccess
//   "User <USER> is currently editing Z_REPRO"
//   T100KEY EU 510

// Equivalent failure via the higher-level write path:
_, err = client.SetSource(ctx,
    "/sap/bc/adt/programs/programs/z_repro",
    "REPORT z_repro.\nWRITE 'hi'.",
    "<any handle>", "", "")
// → 423 com.sap.adt/ExceptionResourceInvalidLockHandle
\`\`\`

The "user is currently editing" message is misleading — it's the same user (and the same \`Client\` instance) that just created the object one call earlier.

## SM12 evidence (S/4)

The orphan enqueue immediately after \`CreateObject\`:

| Field | Value |
|---|---|
| Sperrobjekt | **\`ESRDIRE\`** |
| Tabellenname | \`TRDIR\` |
| Sperrargument | \`Z_REPRO\` |
| Sperrmodus | \`X\` (exclusive) |
[...]

PR #41, opened 2026-04-10
refactor(#35): discovery-first content negotiation for source ops + ATC

### LINKED ISSUE #12: RunATCCheck returns HTTP 500 with empty body on R/3 even with explicit check_variant=DEFAULT
## Summary

`adt.Client.RunATCCheck` against R/3 returns **HTTP 500 with no body**, regardless of which `check_variant` is supplied — including the system default variant that `GetATCCustomizing` reports for the same system. The same call against S/4 works (returns clean results).

This may be an adtler bug (wrong request shape for R/3), an SAP-side configuration issue, or both. I'm filing it here because the failure surfaces through the adtler call boundary and the consumer's wrapper does no transformation; the adtler maintainer is best placed to localize.

## Reproduction (R/3, 2 of 2)

```go
ctx := context.Background()
client := adt.NewClient()

// Without check_variant
result, err := client.RunATCCheck(ctx,
    []string{"/sap/bc/adt/programs/programs/z_bug_repro_20260408_a"}, "")
// → SAP ADT error 500: 

// With explicit DEFAULT (the variant GetATCCustomizing returns for this system)
result, err = client.RunATCCheck(ctx,
    []string{"/sap/bc/adt/programs/programs/z_bug_repro_20260408_a"}, "DEFAULT")
// → identical 500, identical empty body

cust, _ := client.GetATCCustomizing(ctx)
// cust.SystemCheckVariant == "DEFAULT"   ← matches what we passed
```

## Cross-system contrast — S/4 works

```go
result, err := client.RunATCCheck(ctx,
    []string{"/sap/bc/adt/programs/programs/zz_fun_demo"}, "")
// → {WorklistID: "00000000000000000000000000000000", Findings: nil}
```

S/4 returns a valid (empty findings) result without needing `check_variant`. The variant for S/4 is `ZCB_CLEAN_ABAP_1`, not `DEFAULT`.

## Reopened 2026-04-18 — not actually fixed by v0.1.4

The discovery-first negotiation refactor in v0.1.4 (adtler#35 / adtler#41) does **not** fix this bug on R/3 against existing activated programs. Retested with clean mcp-server-abap subprocess smoke tests (isolated binary, `--tools=all`, no preceding customizing call):

| System | Program | v0.1.3 | v0.1.4 |
| --- | --- | --- | --- |
| S/4 `s4u` | `/sap/bc/adt/programs/programs/zz_fun_demo` | ❌ 500 | ✅ ok |
| R/3 `hfq` | `/sap/bc/adt/programs/programs/z_adt_mcp_test_report` | ❌ 500 (empty) | ❌ 500 (empty) |

So v0.1.4 **does** repair S/4 `RunATCCheck` against existing activated programs (nice side effect — the v0.1.3 baseline was also broken there), but R/3 still returns an empty-body 500 with both empty and `DEFAULT` variant.

Why the earlier closure was premature: the integration test `TestRunATCCheck_DiscoveryDriven_Integration` targets a freshly created `$TMP` program, which apparently takes a different R/3 server-side code path than ATC against an existing activated program in a transport-managed package. The test being green is not evidence that the production usage pattern works.

## What I know vs what I'm inferring

**Known facts:**
- 2/2 failures on R/3 (v0.1.3 and v0.1.4), 1/1 success on S/4 v0.1.4 (with clean session), same call shape, same user.
- The 500 has no error body — adtler reports `SAP ADT error 500:` literally with nothing after the colon, so SAP gave us no diagnostic.
- The variant name being passed is the one ADT itself reports as the system default for that R/3 system.
- mcp-server-abap's wrapper (`tools/atc.go:48`) is a one-line passthrough — `client.RunATCCheck(ctx, uris, checkVariant)` — so the failure is at or below adtler's API boundary.
- Discovery-driven content negotiation (landed in v0.1.4) does not change the R/3 outcome, so the root cause is not the outbound `Content-Type`.

**Not yet known:**
- Whether the R/3 ATC handler expects a different request body format than S/4 (worklist setup, check variant ID encoding, etc.).
- Whether the R/3 system has a server-side ATC framework misconfiguration that would surface as a 500 even from Eclipse ADT.
- Whether the empty 500 body is the SAP server failing before it can produce an error document, or whether adtler is dropping the response body.
- Whether ATC against a `$TMP` program vs an existing package-bound activated program takes meaningfully different code paths on R/3 (the integration test passes on `$TMP`, real usage fails on activated programs).

Useful next steps: capture the raw HTTP request adtler sends on R/3 vs S/4 and diff them; try the same flow from Eclipse ADT against the R/3 system to see if it also 500s; extend the adtler integration test to also target an existing activated program (not only freshly created `$TMP`), which would have caught the false green earlier.

## CLAUDE.md note

mcp-server-abap's CLAUDE.md says "On ECC systems this may be required to avoid a server error" referring to `check_variant`. The current observation is that supplying `check_variant=DEFAULT` (the system default) is **not enough** to avoid the 500 — that doc note is incomplete.

## Cross-references

- Consumer-side report: Hochfrequenz/mcp-server-abap#288
- Discovery refactor that did not fix this: adtler#35, adtler#41
- Consumer bump that surfaced the reopening evidence: Hochfrequenz/mcp-server-abap#306

hf-mrdachner:
## Summary

Wire the ADT discovery cache into source operations and `RunATCCheck`, so Accept/Content-Type headers are driven by what the server advertises. Backwards-compatible: all new behaviour falls back to today's hardcoded values when discovery has no entry for the endpoint.

- New `sourceContentType(objectURI)` helper: longest-prefix match against the live discovery cache, picks a preferred type, falls back to `text/plain`.
- New `ensureCSRF(ctx)` helper: unified CSRF + discovery preflight. Replaces inline lock+fetch blocks in `doReadWith`, `doMutateWith`, and is called at the top of source-op entry points so `sourceContentType` sees a populated discovery cache.
- `GetSource`, `GetIncludeSource`, `setSourceWithLockHeader`, `setSourceWithLockParam`, `SetIncludeSource` now consult `sourceContentType` for Content-Type / Accept headers.
- `RunATCCheck` now uses `NegotiateContentType("/sap/bc/adt/atc/runs", [vendor, xml], "application/xml")` for the POST body Content-Type. Suspected fix for #12.
- `SetSource`'s 412 ETag charset retry (from #15) is byte-identical — only the literals were substituted with package constants `contentTypeTextPlain` / `contentTypeTextPlainUTF8`.
- The R/3 vs S/4 lock-handle header-first + query-param retry in `trySetSource` is unchanged.

Closes #35. May close #12 pending the R/3 integration test result.

## Behavior change to watch

On systems where discovery has no entry for the write endpoint, PUT Content-Type now falls back to bare `text/plain` (previously hardcoded `text/plain; charset=utf-8`). If this triggers a 412 on a system that relied on the charset qualifier, the existing #15 retry path catches it and re-sends with the patched ETag — verified by the untouched `source_etag_charset_integration_test.go`. Still, the integration-test agent should pay attention to the SetSource round-trip on both R/3 and S/4.

## Test plan

- [x] `go test ./...` (unit)
- [x] `go build -tags integration ./adt/...`
- [x] `go vet -tags integration ./adt/...`
- [x] `go vet ./...` and `go build ./...`
- [ ] `golangci-lint` (runs in CI — local install has an unrelated go-critic plugin panic)
- [ ] Integration: `TestSourceOperations_DiscoveryDriven_Integration` on R/3 + S/4
- [ ] Integration: `TestRunATCCheck_DiscoveryDriven_Integration` on R/3 + S/4 (may close #12)

## Related

- Design doc: `docs/superpowers/specs/2026-04-10-issue-35-discovery-content-negotiation-design.md`
- Implementation plan: `docs/superpowers/plans/2026-04-10-issue-35-discovery-content-negotiation.md`

### DIFF
--- a/adt/atc.go
+++ b/adt/atc.go
@@ -73,11 +73,17 @@ func (c *httpClient) RunATCCheck(ctx context.Context, objectURIs []string, check
 
 	// Step 1: Trigger ATC run with a worklist to collect results.
 	const worklistID = "0000000000"
+	if err := c.ensureCSRF(ctx); err != nil {
+		return nil, fmt.Errorf("RunATCCheck: %w", err)
+	}
+	ct := c.NegotiateContentType("/sap/bc/adt/atc/runs",
+		[]string{"application/vnd.sap.adt.atc.runs.v1+xml", "application/xml"},
+		"application/xml")
 	resp, err := c.doMutate(ctx, http.MethodPost,
 		"/sap/bc/adt/atc/runs?clientWait=false&worklistId="+worklistID,
 		strings.NewReader(body),
 		map[string]string{
-			"Content-Type": "application/xml",
+			"Content-Type": ct,
 			"Accept":       "application/xml",
 		},
 	)
--- a/adt/atc_discovery_integration_test.go
+++ b/adt/atc_discovery_integration_test.go
@@ -1,0 +1,73 @@
+//go:build integration
+
+package adt_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+)
+
+// TestRunATCCheck_DiscoveryDriven_Integration verifies that after the
+// discovery-first content-negotiation refactor (adtler#35), RunATCCheck
+// still succeeds end-to-end against both R/3 and S/4.
+//
+// Historically RunATCCheck on R/3 returned HTTP 500 because the client sent
+// a hardcoded Content-Type that R/3 did not accept (adtler#12). With the
+// discovery-driven transport, the Content-Type is negotiated from the ATC
+// collection's advertised accept types, which should make the call succeed
+// on R/3 as well as S/4.
+//
+// Uses eachSystem so a single test function validates behaviour against
+// every whitelisted SAP system (R/3 and S/4) in one run. On R/3 this may
+// close adtler#12; on S/4 it is a regression guard.
+func TestRunATCCheck_DiscoveryDriven_Integration(t *testing.T) {
+	ctx := context.Background()
+	for _, sys := range eachSystem(t) {
+		sys := sys
+		t.Run(sys.Name, func(t *testing.T) {
+			// Unique name per system + per run to avoid collisions with
+			// leftover objects from earlier aborted runs.
+			name := fmt.Sprintf("Z_ADT_MCP_ATC_%d", time.Now().UnixNano()%100000)
+			uri := "/sap/bc/adt/programs/programs/" + name
+
+			// 1. Create a disposable $TMP program to point ATC at.
+			if err := sys.Client.CreateObject(ctx, "PROG", name, "$TMP",
+				"adtler#35 discovery-driven ATC integration test", ""); err != nil {
+				t.Fatalf("[%s] CreateObject: %v", sys.Name, err)
+			}
+			t.Logf("[%s] created %s", sys.Name, name)
+
+			// Best-effort cleanup — re-lock then delete. Uses a fresh context
+			// so cleanup still runs if the test context is cancelled.
+			t.Cleanup(func() {
+				lh, lockErr := sys.Client.LockObject(context.Background(), uri)
+				if lockErr != nil {
+					t.Logf("[%s] cleanup lock failed: %v", sys.Name, lockErr)
+					return
+				}
+				if delErr := sys.Client.DeleteObject(context.Background(), uri, lh, ""); delErr != nil {
+					t.Logf("[%s] cleanup delete failed: %v", sys.Name, delErr)
+				}
+			})
+
+			// 2. RunATCCheck — exercises the discovery-driven Content-Type
+			//    negotiation for the /sap/bc/adt/atc/runs POST. Any regression
+			//    or R/3-specific MIME mismatch will surface as HTTP 500/415.
+			result, err := sys.Client.RunATCCheck(ctx, []string{uri}, "DEFAULT")
+			if err != nil {
+				t.Fatalf("[%s] RunATCCheck: %v", sys.Name, err)
+			}
+			if result == nil {
+				t.Fatalf("[%s] RunATCCheck returned nil result", sys.Name)
+			}
+
+			// ATC results are variant-dependent — don't assert a count, just
+			// log it so the test output shows the discovery-driven call
+			// really reached SAP and returned a parsed worklist.
+			t.Logf("[%s] RunATCCheck OK: worklist=%q, %d findings",
+				sys.Name, result.WorklistID, len(result.Findings))
+		})
+	}
+}
--- a/adt/atc_test.go
+++ b/adt/atc_test.go
@@ -1,0 +1,92 @@
+package adt_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/Hochfrequenz/adtler/adt"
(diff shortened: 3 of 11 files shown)


hf-mrdachner:
## Integration test results — R/3 (hfq) + S/4 (s4u)

Ran the two new tests from this PR against both systems via `eachSystem(t)`.

### `TestSourceOperations_DiscoveryDriven_Integration`

| System | Result |
|---|---|
| hfq (ECC / R/3) | ✅ **PASS** — full Lock → GetSource → SetSource → Unlock cycle OK |
| s4u (S/4HANA)  | ✅ **PASS** — full Lock → GetSource → SetSource → Unlock cycle OK |

The discovery-driven content negotiation refactor works end-to-end against both systems. No regression on Get/SetSource for PROG objects.

### `TestRunATCCheck_DiscoveryDriven_Integration`

| System | Result |
|---|---|
| hfq (ECC / R/3) | ❌ **FAIL** — `SAP ADT error 500` (empty body) — **same symptom as #12** |
| s4u (S/4HANA)  | ✅ **PASS** — worklist returned successfully, 0 findings |

**Conclusion for #12:** The hypothesis that R/3 rejected the hardcoded `application/xml` Content-Type is **not confirmed**. Discovery on R/3 does not advertise a vendor MIME type for `/sap/bc/adt/atc/runs`, so `NegotiateContentType` falls back to `application/xml` — i.e. the exact same Content-Type the pre-refactor code sent. The HTTP 500 persists.

R/3's ATC 500 must therefore be caused by something other than Content-Type negotiation (SAP-side configuration, missing ATC variant, authorization, or something else — the 500 body is empty). #12 stays open with `blocked:sap-investigation`.

**Proposal:** keep the ATC integration test in the PR as a regression guard for S/4 and as a future closure-verifier for #12, but add a per-system skip for `hfq` pointing at #12 so integration runs don't fail on a known-open issue. Alternatively, leave the test failing as a loud reminder. Your call.

### Cleanup notes

Both test cleanups (`t.Cleanup` DeleteObject) failed with `ExceptionResourceNoAccess: "User DACHNERM is currently editing Z_ADT_MCP_*"` for the freshly-created probe objects. The actual test assertions passed; this is a post-cycle enqueue that didn't clear in time. Leftover objects in `$TMP`:

- `Z_ADT_MCP_DISC_62400` (s4u)
- `Z_ADT_MCP_ATC_80100` (s4u)

No leftovers on hfq. Will clean up manually.

### Environment

- Branch: `refactor/35-discovery-content-negotiation` at commit `bb407db`
- Command: `SAP_INTEGRATION_SYSTEMS="hfq,s4u" go test -tags=integration -v -run 'TestSourceOperations_DiscoveryDriven_Integration|TestRunATCCheck_DiscoveryDriven_Integration' ./adt/...`
- Duration: 20.19s

hf-mrdachner:
## Cleanup update

Attempted to delete the two leftover probe objects on s4u (`Z_ADT_MCP_DISC_62400`, `Z_ADT_MCP_ATC_80100`) via the sap-adt MCP, but both are held by a session-bound ENQUEUE lock ("User DACHNERM is currently editing ...") that's not in the current session's lock map. This is the same pattern as #4 (ESRDIRE enqueue left behind by `CreateObject` on S/4) but for an object-level enqueue instead. Since the Go test session that created the lock has already exited, the enqueue will clear on SAP-side session timeout. Until then, the objects are orphaned in `$TMP` and don't block anything.

Noting this as a known loose end, not a merge blocker.

Issue #43, opened 2026-04-14
feat: extract namespace+type from <exc:exception> envelope into ADTError


hf-kklein:
## Summary

When SAP returns an ADT exception, the envelope carries a stable, language-independent identifier — namespace + type ID — that today is parsed away. `parseADTError` in `adt/client.go` only extracts the human-readable `<message>` and discards `<namespace id="…"/>` and `<type id="…"/>`.

This identifier is the SAP-ADT equivalent of an ABAP MSGID/MSGNO: stable across SAP releases and locales, and ideal for downstream correlation, filtering, and conditional retry logic.

## Current state

The real envelope shape (verified in `adt/object_test.go:280-285`) is:

```xml
<exc:exception xmlns:exc="http://www.sap.com/abapxml/types/communicationframework">
  <namespace id="com.sap.adt"/>
  <type id="ExceptionResourceWrongData"/>
  <message lang="EN">Resource ZCL_TEST: wrong input data for processing</message>
</exc:exception>
```

`parseADTError` (`adt/client.go:494`) currently parses the older `<ExceptionText><message>` form into:

```go
type ADTError struct {
    StatusCode int
    Message    string
}
```

The newer `<exc:exception>` form mostly works because `<message>` happens to be picked up too, but `namespace` and `type` are dropped. Callers that want to act on a specific exception class (e.g. `ExceptionResourceLocked` for a 423) have to substring-match on the localised message — fragile and locale-dependent.

## Proposed change

1. Extend `ADTError`:

   ```go
   type ADTError struct {
       StatusCode int
       Namespace  string  // e.g. "com.sap.adt"  — empty if unknown
       Type       string  // e.g. "ExceptionResourceLocked"  — empty if unknown
       Message    string
   }
   ```

2. Update `parseADTError` to handle both schemas:
   - `<exc:exception>` with `<namespace id>` / `<type id>` / `<message>`
   - `<ExceptionText><message>` (legacy, keep working — `Namespace`/`Type` stay empty)

3. Update `Error()` to include the type ID when present:
   - With type:    `SAP ADT error 423 (ExceptionResourceLocked): Object is locked by user X`
   - Without type: `SAP ADT error 500: Internal server error` (current format preserved)

## Backwards compatibility

- `ADTError.Message` continues to carry exactly the same string it carries today.
- New fields default to `""`, so existing code reading `adtErr.Message` is unaffected.
[...]

PR #47, opened 2026-04-27
feat(#43): extract namespace+type from <exc:exception> into ADTError

hf-mrdachner:
## Summary

- Adds `Namespace` and `Type` fields to `ADTError`, populated from `<exc:exception>` envelopes (issue #43).
- Adds four exported `ExceptionType*` constants for the exception IDs adtler reacts to internally.
- Rewrites `isInvalidLockHandle` and `isPreconditionFailed` to prefer the structured `Type` field, with status-code fallback for legacy `<ExceptionText>` responses.
- `Error()` includes the type ID in parentheses when present; format unchanged otherwise.

## Behaviour change

A 423 `ExceptionResourceLocked` (object held by another user) no longer triggers the lock-handle-delivery retry — that retry is for `ExceptionResourceInvalidLockHandle` only. The previous "any 423" check was over-broad. Existing legacy responses (Type empty) keep the status-code-only behaviour.

## Consumer impact

Consumers that prefix-match \`"SAP ADT error 423:"\` on \`Error()\` output will see the new \`(ExceptionResourceLocked)\` qualifier when SAP returns an \`<exc:exception>\` envelope. The follow-up bump in \`mcp-server-abap\` (Hochfrequenz/mcp-server-abap#310) will update the canonical event logger to consume \`adtErr.Type\` directly.

## Test plan

- [x] Unit: 13 new tests in \`adt/client_internal_test.go\` covering envelope parsing, edge cases, \`Error()\` formatting, and predicate behaviour (incl. shared \`runPredicateTests\` helper).
- [x] Unit: existing \`TestParseADTError_*\` and \`TestDeleteObject_PropagatesETagFetchError\` stay green.
- [x] Local integration: \`TestADTError_NamespaceTypeExtraction_MultiSystem\` PASS on hfq with \`Namespace="com.sap.adt"\`, \`Type="ExceptionResourceNotFound"\`.
- [ ] Multi-system integration: pending integration agent (R/3 + S/4).

## Cross-references

- Issue: Hochfrequenz/adtler#43
- Consumer-side: Hochfrequenz/mcp-server-abap#310
- Design spec: \`docs/superpowers/specs/2026-04-27-issue-43-adt-error-namespace-type-design.md\`
- Implementation plan: \`docs/superpowers/plans/2026-04-27-issue-43-adt-error-namespace-type.md\`


hf-mrdachner:
## Independent Review — PR #47

*Reviewer: Claude Sonnet 4.6 via superpowers:code-reviewer (no prior conversation context).*

---

### Protocol followed

- Read PR via `gh pr view 47` and `gh pr diff 47`
- Read adtler issue #43 for context
- Attempted consumer-side issue mcp-server-abap#310 — accessible; it is about build version / request_id in the canonical log event, not directly about ADT exception types. The PR description cross-references it for context only; no review impact.
- Created dedicated worktree at `C:/Users/dachner/AppData/Local/Temp/adtler-review-pr47` from `remotes/origin/feat/43-adt-error-namespace-type`
- Ran `go test ./...` — **all green** (adt, adt/adtxml, adt/custexport, auth)
- Ran `go build -tags integration ./adt/...` — **no output (clean)**
- Ran `go vet -tags integration ./adt/...` — **no output (clean)**
- Worktree removed after review

---

### Strengths

- The four-layer fallback strategy is well-documented and each layer is clearly responsible for one format. The refactor is non-breaking for all three legacy formats (ExceptionText XML, SAP HTML page, plain text).
- `xml:"exception"` on the anonymous struct correctly matches the local name of `<exc:exception>` regardless of prefix — Go's `encoding/xml` strips prefixes on decode. Verified by running a test binary against the actual fixture shapes from `object_test.go` and `refactoring_test.go`.
- The guard `excEnv.Message != ""` is the right heuristic for "layer 1 owns this body". It mirrors the existing guard on the legacy layer and prevents silent false-positive matches.
- `isInvalidLockHandle` / `isPreconditionFailed` rewrite is correct and the behaviour-change callout in the PR description is accurate: a structured 423 `ExceptionResourceLocked` (object held by another user) no longer triggers the lock-handle retry. The legacy fallback (`Type == ""` → status code check) preserves old behaviour for R/3 responses that do not yet carry structured envelopes.
- `ExceptionType*` constants are exported, well-documented, and placed where they are easiest to find by consumers.
- Table-driven predicate tests with the shared `runPredicateTests` helper are clean, DRY, and cover the nil case.
- Integration test uses `eachSystem(t)`, correct build tag (`//go:build integration`), correct package (`adt_test`), and the non-existent URI trick is a reliable cross-system trigger. Not asserting exact Type IDs to handle R/3 / S/4 variation is explicitly justified in the comment.
- Commit history is disciplined: TDD order (test→fail→impl→green) is reflected in the commit log.

---

### Findings

**Important: Layer 1 matches `<exc:exception>` regardless of namespace URI**

Go's `encoding/xml` matches element names by local name only when no namespace is given in the struct tag. The struct tag is `xml:"exception"`, with no namespace URI. This means layer 1 will also claim bodies from a *different* SAP exception namespace — specifically the `http://www.sap.com/adt/exceptions` URI used in the three refactoring test mocks in `adt/refactoring_test.go` (lines 225, 261, 305). In those fixtures the child elements are `<exc:message>` (prefixed), which also matches `xml:"message"` by local name — so `excEnv.Message` is non-empty and layer 1 claims the body.

Concretely: `TestRenameEvaluateError`, `TestRenamePreviewError`, and `TestRenameExecuteError` now route through layer 1 instead of layer 2. All three still pass today because those tests only assert `strings.Contains(err.Error(), "Rename …")` on the *wrapping* error, not on the underlying ADT error format. However, the `Namespace` field on those errors will be `""` (the fixture has no `<namespace>` child) and `Type` will be `""`, so an observant caller checking `adtErr.Type` on a rename failure would see an empty string even though the response *was* structured XML — just from a different SAP namespace. This is an undocumented side-effect of the local-name-only match.

This is categorised as **Important** (not Critical) because: (a) the affected tests pass, (b) `Namespace` and `Type` remaining empty for those errors is arguably correct given the different NS URI, and (c) the rename wrappers don't surface ADTError fields to callers. A follow-up should either add the namespace URI to the struct tag (`xml:"http://www.sap.com/abapxml/types/communicationframework exception"`) to narrow the match, or document the local-name-only behaviour explicitly so future fixture authors are aware.

**Minor: `TestParseADTError_ExcExceptionMissingMessage` exercises a degenerate case with a subtle design quirk**

When `<exc:exception>` is present but has no `<message>`, layer 1 does not claim the body (correct). The body then falls through to layer 4 (plain text), returning raw XML in `Message` with empty `Namespace` and `Type`. The test asserts this behaviour and the comment explains it. However, this means a SAP system that sends a structurally valid `<exc:exception>` with namespace+type but no message text would produce an `ADTError` with the raw XML as `Message` and empty `Namespace`/`Type` — losing the structured identity. In practice this is unlikely given that SAP always includes `<message>`, but the design note in the doc-comment would benefit from explicitly flagging this edge case.

**Minor: Consumer-side `Error()` format break is acknowledged but no migration path is specified**

The PR description correctly notes the format change (`"SAP ADT error 423:"` → `"SAP ADT error 423 (ExceptionResourceLocked):"`). The consumer bump in mcp-server-abap#310 is referenced but that issue is actually about a different feature (build version + request_id). The actual consumer migration for the changed `Error()` format should be tracked explicitly — either as a follow-up task or a linked issue — to ensure it does not get lost.

---

### Verdict: GO with notes

The implementation is correct, well-tested, and the idiomatic Go patterns (anonymous struct for one-shot XML decode, `errors.As` for predicate functions, table-driven tests, `eachSystem` integration pattern) are all sound. All unit tests pass cleanly and the integration test is appropriately structured. The behaviour-change callout in the PR description is accurate and important.

The two notes above — namespace-URI specificity in the struct tag, and the consumer format migration tracking — are improvements worth considering but do not block merge. The codebase is in a better state with this PR than without it.

hf-mrdachner:
## Integration test result — PR #47

`SAP_INTEGRATION_SYSTEMS=hfq,s4u go test -tags=integration -v -run TestADTError_NamespaceTypeExtraction_MultiSystem ./adt/...`

| System | Result | StatusCode | Namespace | Type | Message language |
|---|---|---|---|---|---|
| **R/3** (`hfq`) | ✅ PASS | 404 | `com.sap.adt` | `ExceptionResourceNotFound` | German |
| **S/4** (`s4u`) | ✅ PASS | 404 | `com.sap.adt` | `ExceptionResourceNotFound` | English |

Both systems return identical `Namespace` + `Type` for the same trigger (`LockObject` on a non-existent class URI). Only the localised `Message` text differs — which is exactly the locale-dependence problem this PR fixes by exposing the locale-stable `Type` field.

```
=== RUN   TestADTError_NamespaceTypeExtraction_MultiSystem
=== RUN   TestADTError_NamespaceTypeExtraction_MultiSystem/hfq
    [hfq] StatusCode=404 Namespace="com.sap.adt" Type="ExceptionResourceNotFound" Message="CLASS ZCL_ADT_MCP_EXC_TEST_XXX ist nicht vorhanden"
=== RUN   TestADTError_NamespaceTypeExtraction_MultiSystem/s4u
    [s4u] StatusCode=404 Namespace="com.sap.adt" Type="ExceptionResourceNotFound" Message="CLASS ZCL_ADT_MCP_EXC_TEST_XXX does not exist"
--- PASS: TestADTError_NamespaceTypeExtraction_MultiSystem (1.04s)
    --- PASS: TestADTError_NamespaceTypeExtraction_MultiSystem/hfq (0.25s)
    --- PASS: TestADTError_NamespaceTypeExtraction_MultiSystem/s4u (0.79s)
```

**Cleanup notes:** The test triggers a 404 by attempting to lock a class URI that does not exist; no objects created on either system, no transports involved, nothing to clean up.

**Empirical observation:** the `<exc:exception>` schema is consistent across R/3 and S/4 for the LockObject-on-missing-object trigger, both producing `com.sap.adt` / `ExceptionResourceNotFound`. This validates the design's assumption that `Type` is a locale- and release-stable identifier.

hf-mrdachner:
## Review notes addressed (commit `ed0de49`)

Following the independent review (GO with notes), all three findings have been addressed:

**1. Important — Layer 1 namespace specificity** (now fixed)

`parseADTError` layer 1 still claims `<exc:exception>` by local-name match — but `Namespace`/`Type` are now populated only when the root element's XML namespace URI is `http://www.sap.com/abapxml/types/communicationframework`. For the older SAP namespace (`http://www.sap.com/adt/exceptions`, used by refactoring endpoints), the body still matches layer 1 and `<message>` is still extracted into `Message`, but `Namespace` and `Type` correctly stay empty (the older schema does not carry those children). Behaviour is preserved for the rename test fixtures (`TestRenameEvaluateError`, `TestRenamePreviewError`, `TestRenameExecuteError` all green).

Added regression test: `TestParseADTError_ExcExceptionOldNamespace`.

**2. Minor — Missing-message fall-through edge case** (now documented)

Added a paragraph to `parseADTError`'s doc comment flagging that an `<exc:exception>` body without a `<message>` child falls through to later layers, losing any structured `<namespace>`/`<type>` identifiers. Behaviour unchanged; just made explicit.

**3. Minor — Consumer-side `Error()` migration tracker**

Filed Hochfrequenz/mcp-server-abap#363 to track the consumer-side migration off `Error()`-prefix-matching to structural `adtErr.Type` checks. The reviewer correctly noted that mcp-server-abap#310 (referenced in the original PR description) is about a different feature.

## Local verification

- `go test ./...` — green
- `go test -run TestParseADTError -v ./adt/...` — 10/10 pass (4 pre-existing + 5 from this PR + 1 new old-namespace regression test)
- `go test -run TestRename ./adt/...` — green
- `golangci-lint --enable dupl,goconst,gocyclo` — 0 issues

PR #50, opened 2026-04-27
fix(#12): adopt 3-step ATC flow to fix R/3 500

### LINKED ISSUE #12: RunATCCheck returns HTTP 500 with empty body on R/3 even with explicit check_variant=DEFAULT

hf-mrdachner:
## Summary

Closes #12. Consumer-side: Hochfrequenz/mcp-server-abap#288.

`RunATCCheck` previously used a 2-step shortcut — it posted to `/sap/bc/adt/atc/runs` with a placeholder `worklistId="0000000000"` and put `checkVariant` in the body. This worked on S/4 (silent auto-create) but returned HTTP 500 with empty body on R/3, which strictly requires a real worklist to exist before `/runs` is invoked.

This PR adopts the SAP ADT 3-step worklist flow — the shape was observed in the [abap-adt-api](https://github.com/marcellourbani/abap-adt-api) reference implementation (`src/api/atc.ts`) during investigation and verified end-to-end against real R/3 and S/4 systems. No code was copied; the Go implementation is independently written from observation of the wire protocol.

1. `POST /sap/bc/adt/atc/worklists?checkVariant={variant}` — `Accept: text/plain`, response body is the SAP-assigned worklist ID.
2. `POST /sap/bc/adt/atc/runs?worklistId={obtainedID}` — body no longer carries `checkVariant`. If the response carries a different ID in `<worklistRun>`, that ID supersedes the input.
3. `GET /sap/bc/adt/atc/worklists/{id}` — findings, unchanged shape.

Discovery-driven Content-Type negotiation for the `/runs` POST (adtler#35) is preserved.

## Why the previous attempt didn't fix this

PR #41 hypothesised that R/3 was rejecting the hardcoded `application/xml` Content-Type and tried to fix it via `NegotiateContentType`. As @hf-mrdachner reported on that PR, R/3's discovery doesn't advertise a vendor MIME for `/atc/runs`, so the negotiation falls back to the same `application/xml` the pre-refactor code sent — leaving the 500 in place. Root cause is the missing worklist creation step, not the Content-Type. The empty 500 body is consistent with R/3's ATC handler crashing before it can produce a diagnostic when handed a non-existent worklist ID.

## Behavior change

- Empty `checkVariant` now defaults to `"DEFAULT"` (R/3's system default). Previously the `checkVariant` attribute was simply omitted from the body. **S/4 callers should pass their site-specific variant explicitly** (e.g. `ZCB_CLEAN_ABAP_1`) — `DEFAULT` may produce different findings on S/4 than the system default.
- The deliberate choice **not** to auto-call `GetATCCustomizing` for an empty variant: that call triggers adtler#44 (session-state issue) on S/4 and adds an extra round-trip. Callers who need to discover their system variant can call `GetATCCustomizing` themselves.

## Test plan

- [x] `go test ./...` (unit) — all green; 8 new ATC tests cover call ordering, query strings, accept header, body shape (no `checkVariant`), worklist ID propagation through both responses, the `DEFAULT` fallback, and URL-encoding of variant names.
- [x] `go vet ./...` and `go vet -tags integration ./adt/...` — clean
- [x] `go build ./...` and `go build -tags integration ./adt/...` — clean
- [x] `golangci-lint run --enable dupl,goconst,gocyclo ./...` — 0 issues
- [x] `gofmt -l .` — clean
- [x] Integration: `TestRunATCCheck_DiscoveryDriven_Integration` on R/3 + S/4 — both PASS (results posted as PR comment)
- [x] Manual probe against an activated package-bound program on R/3 + S/4 (the production failure path of #12) — both PASS (results posted as PR comment)

## Files

- `adt/atc.go` — split `RunATCCheck` into `createATCWorklist` / `triggerATCRun` / `fetchATCWorklist` helpers; add `defaultATCCheckVariant` constant; reuse the existing `adtxml.ATCWorklistRun` struct (was declared but unused).
- `adt/atc_test.go` — 8 new tests + an `atcMockCapture` helper for the 3-step flow; existing Content-Type tests adapted to the new flow.
- `adt/atc_discovery_integration_test.go` — docstring updated; test surface unchanged (still creates a `$TMP` program, calls `RunATCCheck`, asserts no error).

### DIFF
--- a/adt/atc.go
+++ b/adt/atc.go
@@ -66,8 +66,9 @@ func (c *httpClient) GetATCCustomizing(ctx context.Context) (*ATCCustomizingResu
 //
 // The earlier 2-step shortcut (POST /runs with worklistId="0000000000",
 // checkVariant in body) returned HTTP 500 with empty body on R/3 — see
-// adtler#12. The 3-step flow is the one Eclipse ADT uses and works on
-// both R/3 and S/4.
+// adtler#12. The 3-step shape was observed in the abap-adt-api reference
+// implementation (src/api/atc.ts) and verified against the SAP ADT REST
+// API on both R/3 and S/4.
 //
 // If checkVariant is empty, "DEFAULT" is used. Callers on S/4 should pass
 // their site-specific variant explicitly.
@@ -130,7 +131,8 @@ func (c *httpClient) createATCWorklist(ctx context.Context, checkVariant string)
 // triggerATCRun performs step 2: posts the object references to the runs
 // endpoint, scoped to worklistID. SAP echoes back the worklist ID (and a
 // timestamp) in a <worklistRun> envelope; if the response carries a
-// non-empty ID, that ID supersedes the input.
+// non-empty ID, that ID supersedes the input (same supersession rule
+// observed in abap-adt-api).
 func (c *httpClient) triggerATCRun(ctx context.Context, worklistID string, objectURIs []string) (string, error) {
 	body := buildATCRunBody(objectURIs)
 
--- a/adt/atc.go
+++ b/adt/atc.go
@@ -6,10 +6,18 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/Hochfrequenz/adtler/adt/adtxml"
 )
+
+// defaultATCCheckVariant is used when the caller passes an empty variant.
+// SAP R/3's standard ATC check variant is "DEFAULT" (also reported by
+// GetATCCustomizing on R/3). S/4 callers should pass their site-specific
+// variant explicitly (often "ZCB_CLEAN_ABAP_1") — DEFAULT may produce
+// different findings on S/4 than the system default.
+const defaultATCCheckVariant = "DEFAULT"
 
 func (c *httpClient) GetATCCustomizing(ctx context.Context) (*ATCCustomizingResult, error) {
 	resp, err := c.doRead(ctx, "/sap/bc/adt/atc/customizing", map[string]string{
@@ -45,42 +53,95 @@ func (c *httpClient) GetATCCustomizing(ctx context.Context) (*ATCCustomizingResu
 	return result, nil
 }
 
+// RunATCCheck runs an ATC check against the given object URIs. It implements
+// the canonical 3-step Eclipse ADT flow:
+//
+//  1. POST /sap/bc/adt/atc/worklists?checkVariant={variant} — creates a
+//     worklist seeded with the requested check variant. The response body
+//     is the worklist ID as plain text.
+//  2. POST /sap/bc/adt/atc/runs?worklistId={id} — triggers the actual run.
+//     The body carries the object references but no longer carries
+//     checkVariant (it's already established on the worklist).
+//  3. GET /sap/bc/adt/atc/worklists/{id} — fetches findings.
+//
+// The earlier 2-step shortcut (POST /runs with worklistId="0000000000",
+// checkVariant in body) returned HTTP 500 with empty body on R/3 — see
+// adtler#12. The 3-step flow matches what abap-adt-api / Eclipse ADT do
+// and works on both R/3 and S/4.
+//
+// If checkVariant is empty, "DEFAULT" is used. Callers on S/4 should pass
+// their site-specific variant explicitly.
 func (c *httpClient) RunATCCheck(ctx context.Context, objectURIs []string, checkVariant string) (*ATCResult, error) {
-	// Build object references XML. Escape URIs to prevent XML injection.
-	var refs strings.Builder
-	for _, uri := range objectURIs {
-		var escaped strings.Builder
-		_ = xml.EscapeText(&escaped, []byte(uri))
-		fmt.Fprintf(&refs, `<adtcore:objectReference adtcore:uri="%s"/>`, escaped.String())
+	if checkVariant == "" {
+		checkVariant = defaultATCCheckVariant
 	}
 
-	variantAttr := ""
-	if checkVariant != "" {
-		var escaped strings.Builder
-		_ = xml.EscapeText(&escaped, []byte(checkVariant))
-		variantAttr = fmt.Sprintf(` checkVariant="%s"`, escaped.String())
-	}
-
-	body := fmt.Sprintf(`<?xml version="1.0" encoding="UTF-8"?>`+
-		`<atc:run xmlns:atc="http://www.sap.com/adt/atc" `+
-		`xmlns:adtcore="http://www.sap.com/adt/core" maximumVerdicts="100"%s>`+
-		`<objectSets>`+
-		`<objectSet kind="inclusive">`+
-		`<adtcore:objectReferences>%s</adtcore:objectReferences>`+
-		`</objectSet>`+
-		`</objectSets>`+
-		`</atc:run>`, variantAttr, refs.String())
-
-	// Step 1: Trigger ATC run with a worklist to collect results.
-	const worklistID = "0000000000"
 	if err := c.ensureCSRF(ctx); err != nil {
 		return nil, fmt.Errorf("RunATCCheck: %w", err)
 	}
+
+	// Step 1: Create a worklist for this variant. SAP returns the worklist
+	// ID as plain text in the response body.
+	worklistID, err := c.createATCWorklist(ctx, checkVariant)
+	if err != nil {
+		return nil, fmt.Errorf("RunATCCheck: %w", err)
+	}
+
+	// Step 2: Trigger the run, scoping it to the requested object URIs.
+	worklistID, err = c.triggerATCRun(ctx, worklistID, objectURIs)
+	if err != nil {
+		return nil, fmt.Errorf("RunATCCheck: %w", err)
+	}
+
+	// Step 3: Fetch the resulting findings.
+	return c.fetchATCWorklist(ctx, worklistID)
+}
+
+// createATCWorklist performs step 1 of the ATC flow and returns the
+// SAP-assigned worklist ID. The response body is plain text.
+func (c *httpClient) createATCWorklist(ctx context.Context, checkVariant string) (string, error) {
+	q := url.Values{}
+	q.Set("checkVariant", checkVariant)
+	resp, err := c.doMutate(ctx, http.MethodPost,
+		"/sap/bc/adt/atc/worklists?"+q.Encode(),
+		nil,
+		map[string]string{
+			"Accept": "text/plain",
+		},
+	)
+	if err != nil {
+		return "", fmt.Errorf("create worklist: %w", err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+	if err := checkResponse(resp); err != nil {
+		return "", fmt.Errorf("create worklist: %w", err)
+	}
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", fmt.Errorf("create worklist reading body: %w", err)
+	}
+	id := strings.TrimSpace(string(body))
+	if id == "" {
+		return "", fmt.Errorf("create worklist: SAP returned empty worklist ID")
+	}
+	return id, nil
+}
+
+// triggerATCRun performs step 2: posts the object references to the runs
+// endpoint, scoped to worklistID. SAP echoes back the worklist ID (and a
+// timestamp) in a <worklistRun> envelope; if the response carries a
+// non-empty ID, that ID supersedes the input (matching abap-adt-api).
+func (c *httpClient) triggerATCRun(ctx context.Context, worklistID string, objectURIs []string) (string, error) {
+	body := buildATCRunBody(objectURIs)
+
 	ct := c.NegotiateContentType("/sap/bc/adt/atc/runs",
 		[]string{"application/vnd.sap.adt.atc.runs.v1+xml", "application/xml"},
 		"application/xml")
+
+	q := url.Values{}
+	q.Set("worklistId", worklistID)
 	resp, err := c.doMutate(ctx, http.MethodPost,
-		"/sap/bc/adt/atc/runs?clientWait=false&worklistId="+worklistID,
+		"/sap/bc/adt/atc/runs?"+q.Encode(),
 		strings.NewReader(body),
 		map[string]string{
 			"Content-Type": ct,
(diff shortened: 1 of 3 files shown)


hf-mrdachner:
## Integration test results — R/3 (hfq) + S/4 (s4u)

Ran on this PR's branch (`fix/12-atc-three-step-worklist-flow @ 6819892`) against both systems via `eachSystem(t)`. **Both PASS, including the production failure path (activated package-bound program on R/3) that the previous closure missed.**

### `TestRunATCCheck_DiscoveryDriven_Integration` ($TMP path)

| System | Result |
|---|---|
| hfq (ECC / R/3) | ✅ **PASS** — `worklist="00505694359E1FE190CFD19C1705181C"`, 1 finding |
| s4u (S/4HANA)  | ✅ **PASS** — `worklist="005056BA9BFB1FE190CFD3D5488E1299"`, 1 finding |

Note: SAP now returns real 32-char hex worklist IDs (the previous code's hardcoded 10-zero placeholder is gone — confirms the new flow really is creating worklists via step 1, not relying on the old shortcut).

```
SAP_INTEGRATION_SYSTEMS="hfq,s4u" go test -tags=integration -v \
  -run TestRunATCCheck_DiscoveryDriven_Integration ./adt/...
=== RUN   TestRunATCCheck_DiscoveryDriven_Integration
=== RUN   TestRunATCCheck_DiscoveryDriven_Integration/hfq
    [hfq] created Z_ADT_MCP_ATC_64400
    [hfq] RunATCCheck OK: worklist="00505694359E1FE190CFD19C1705181C", 1 findings
=== RUN   TestRunATCCheck_DiscoveryDriven_Integration/s4u
    [s4u] created Z_ADT_MCP_ATC_85900
    [s4u] RunATCCheck OK: worklist="005056BA9BFB1FE190CFD3D5488E1299", 1 findings
--- PASS: TestRunATCCheck_DiscoveryDriven_Integration (31.27s)
```

### Activated package-bound program path (production failure path of #12)

The issue body explicitly distinguished the `$TMP` path from the production "activated program in transport-managed package" path. I added a temporary probe (`TestRunATCCheck_ActivatedProgram_ProbeOnly`, not committed) that runs ATC against the existing `Z_ADT_MCP_TEST_REPORT` fixture (activated, in package `Z_ADT_MCP_TEST` under `HFQK902857` on hfq).

| System | Result |
|---|---|
| hfq (ECC / R/3) | ✅ **PASS** — `worklist="00505694359E1FE190CFD9AE8641581C"`, 1 finding |
| s4u (S/4HANA)  | ✅ **PASS** — `worklist="005056BA9BFB1FE190CFDA8B4ED53299"`, 1 finding |

```
=== RUN   TestRunATCCheck_ActivatedProgram_ProbeOnly/hfq
    [hfq] RunATCCheck OK on activated /sap/bc/adt/programs/programs/Z_ADT_MCP_TEST_REPORT: worklist="00505694359E1FE190CFD9AE8641581C", 1 findings
=== RUN   TestRunATCCheck_ActivatedProgram_ProbeOnly/s4u
    [s4u] RunATCCheck OK on activated /sap/bc/adt/programs/programs/Z_ADT_MCP_TEST_REPORT: worklist="005056BA9BFB1FE190CFDA8B4ED53299", 1 findings
--- PASS: TestRunATCCheck_ActivatedProgram_ProbeOnly (8.23s)
```

This is the exact failure mode #12 reports — under v0.1.4 this returns HTTP 500 with empty body on R/3. With this PR it succeeds.

Probe file deleted; the committed regression guard remains the existing `TestRunATCCheck_DiscoveryDriven_Integration`.

### Cleanup notes

- `s4u` cleanup of the freshly-created `$TMP` probe object failed with `ExceptionResourceNoAccess: User DACHNERM is currently editing Z_ADT_MCP_ATC_85900` — same orphan-enqueue pattern noted on PR #41 (post-cycle session-bound enqueue that doesn't clear in time). Test assertion still passed; will clean up via SAP-side timeout. Not a merge blocker.
- No leftovers on `hfq`.

### Environment

- Branch: `fix/12-atc-three-step-worklist-flow @ 6819892`
- Command (committed test): `SAP_INTEGRATION_SYSTEMS="hfq,s4u" go test -tags=integration -v -run TestRunATCCheck_DiscoveryDriven_Integration ./adt/...`
- Duration: 31.27s ($TMP test) + 8.23s (activated probe)

Removing `needs:integration-test` shortly.

hf-mrdachner:
## Independent Reviewer Verdict — PR #50

Fresh agent, no prior conversation context. Reviewed PR diff, adtler#12, mcp-server-abap#288, and the changed files in a dedicated worktree. Cross-checked the wire-protocol shape against the [abap-adt-api](https://github.com/marcellourbani/abap-adt-api) reference implementation (`src/api/atc.ts`) since the PR cites it as the source of the 3-step shape.

### Summary

Replaces the broken 2-step shortcut (POST `/runs?worklistId=0000000000` with `checkVariant` in body) with the SAP ADT 3-step worklist flow. Implementation is faithful to the observed reference and CI is green across the board.

### Correctness

Cross-checked against `abap-adt-api/src/api/atc.ts`:

- **Step 1** — abap-adt-api: `POST /sap/bc/adt/atc/worklists?checkVariant=${variant}` with `Accept: text/plain`, returns plain-text body. PR matches exactly (`createATCWorklist`).
- **Step 2** — abap-adt-api: `POST /sap/bc/adt/atc/runs?worklistId=${variant}` with `Accept: application/xml` and `Content-Type: application/xml`, parses `worklistRun/worklistId` from response. PR matches and additionally preserves the discovery-driven Content-Type negotiation from adtler#35 (correct — the Accept advertised by R/3 falls back to `application/xml` anyway, but S/4 vendor-MIME still works).
- **Step 3** — abap-adt-api: `GET /sap/bc/adt/atc/worklists/${runResultId}` with `Accept: application/atc.worklist.v1+xml`. PR matches.

ID propagation through the 3 steps is correct and the step-2 supersession matches reference behaviour. Verified Go's `encoding/xml` happily decodes the namespaced `atc:worklistId` attribute against the unqualified `xml:"worklistId,attr"` tag (also proven end-to-end by `TestRunATCCheck_PrefersWorklistIDFromRunsResponse`).

The fallback-to-step-1-ID when step-2 response is unparseable is a sensible defensive choice — the response is informational; SAP has already kicked off the run by then.

URL encoding via `url.Values.Encode()` is correct and the dedicated test (`TestRunATCCheck_PercentEncodesCheckVariant`) proves it.

### Test coverage

10 unit tests, all pass locally:

- Call ordering (worklists → runs → worklist GET)
- `checkVariant` on the worklists URL query (not in `/runs` body)
- `Accept: text/plain` on worklists POST
- `/runs` body must not contain `checkVariant`
- ID propagation step1→step2→step3
- Step-2 response ID supersedes step-1 ID
- Empty variant defaults to `DEFAULT`
- URL encoding of variant names with special characters
- Discovery-driven Content-Type used on `/runs`
- Fallback Content-Type when discovery is empty

Edge cases I'd consider but won't block on:
- No test for the path where step-2 response body is unparseable (the fallback to step-1 ID). It's a 3-line branch, low risk.
- No test asserting step-2 receives multiple object URIs correctly stitched into one body. The pre-existing `buildATCRunBody` carries that logic unchanged so the gap pre-existed.
- Integration test still uses `$TMP` not a transport-bound activated program — the very gap #12 calls out. The PR author addressed this with an uncommitted probe and integration results posted on the PR show both the `$TMP` path and the activated-program path passing on R/3 + S/4. **Suggestion (non-blocking): consider committing a second integration test against an activated package-bound fixture so this regression guard survives long-term, not just in this cycle.**

### Code quality

- Clean function decomposition: `RunATCCheck` orchestrates, three private helpers (`createATCWorklist`, `triggerATCRun`, `fetchATCWorklist`) each do one thing.
- `buildATCRunBody` extracted as a pure function — testable, small, idiomatic.
- Error wrapping is consistent (`create worklist:`, `trigger run:`, `RunATCCheck fetching worklist:`).
- Doc comments on `RunATCCheck` are excellent — they explain the 3 steps, why the previous shortcut failed, and the `DEFAULT` semantics.
- `defaultATCCheckVariant` const + `atcWorklistsPath` / `atcRunsPath` test consts kill the `goconst` lint smell.
- `t.Cleanup(srv.Close)` in the test mock is a nice modernization over `defer srv.Close()`.

Minor nit (not a blocker): the doc comment on `triggerATCRun` says `<worklistRun>` envelope carries "the worklist ID (and a timestamp)". The `Timestamp` field is parsed into `adtxml.ATCWorklistRun` but unused. Either drop it from the struct or note the deliberate retention. Tiny.

### SAP/ADT semantics

- Accept/Content-Type per step matches abap-adt-api.
- Discovery-driven `Content-Type` for step 2 is preserved (still useful on S/4 where vendor MIME is advertised).
- Empty-variant default to `DEFAULT` rather than calling `GetATCCustomizing` — correct trade-off given adtler#44, and the PR description warns S/4 callers.
- CSRF preflight (`ensureCSRF`) is called once before step 1, then inherited by step 2's `doMutate`. Correct.
- Step 3 uses `doRead` (no CSRF). Correct — it's a GET.
- The `clientWait=false` query param the old code passed to `/runs` is dropped. abap-adt-api doesn't pass it either, so this is consistent.

### Risk / notes

- **Behaviour change**: empty `check_variant` now defaults to `"DEFAULT"`. The PR description explicitly flags this for S/4 callers. Documented.
- **Integration coverage gap**: the committed integration test still targets `$TMP`. Mitigated by the developer's manual probe against an activated package-bound program (results posted on PR), but the long-term regression guard for #12's exact failure path isn't in the test suite. Suggest follow-up issue to add it.
- **`Timestamp` field in `ATCWorklistRun`**: unused. Cosmetic.

### Local verification

In a fresh worktree on `pr-50`:

- `go test ./...` — all packages PASS
- `go build -tags integration ./adt/...` — clean
- `go test -v -run TestRunATCCheck ./adt/...` — all 10 ATC tests PASS

CI per `gh pr checks 50`: lint, test (1.25 + 1.26), coverage, CodeQL, Analyze (go) — all green.

### Workflow adherence

- Branch name `fix/12-atc-three-step-worklist-flow` ✓
- Commit message `fix(#12): adopt canonical 3-step ATC flow to fix R/3 500` ✓
- PR description cross-references both adtler#12 and mcp-server-abap#288 ✓
- Multi-system integration test using `eachSystem(t)` ✓
- Integration-test agent results already posted ✓

### Verdict

**GO**

PR's wire-protocol claims hold up against `abap-adt-api/src/api/atc.ts`, all unit tests pass, lint clean, CI green, real-SAP integration results posted (including the exact production failure path #12 calls out). The two issues I'd raise are both non-blocking follow-ups: (1) commit a second integration test exercising an activated package-bound program so the `#12` regression guard isn't only in the developer's local run; (2) decide whether to drop the unused `Timestamp` field on `ATCWorklistRun`.

PR #54, opened 2026-05-08
fix(#8): pass source body to NavigateToDefinition

### LINKED ISSUE #8: NavigateToDefinition always returns the input source URI as the result
## Summary

\`adt.Client.NavigateToDefinition(ctx, sourceURI)\` always returns the **input \`sourceURI\` byte-for-byte**, regardless of where the cursor is positioned. It never resolves cross-references. Verified across 5 calls on 2 systems (S/4 and R/3) and 2 object types (PROG, CLAS), with cursor positions clearly inside identifiers that have well-known definitions (\`cl_abap_random_int\`, \`create\`, etc.).

## Reproduction (5 of 5)

\`\`\`go
out, err := client.NavigateToDefinition(ctx,
    "/sap/bc/adt/programs/programs/zz_fun_demo/source/main#start=30,16")
// out == "/sap/bc/adt/programs/programs/zz_fun_demo/source/main#start=30,16"
// err == nil
\`\`\`

In \`ZZ_FUN_DEMO\`, line 30 is \`DATA(rnd) = cl_abap_random_int=>create(\` — column 16 is inside the class identifier \`cl_abap_random_int\` (columns 15-32). A correct result should point at \`/sap/bc/adt/oo/classes/cl_abap_random_int\` (or the \`create\` method definition for column 35+), not echo back the source position.

Same identity-mapping behavior at columns 18, 20, and on a different program (\`z_bug_repro_20260408_a\` at line 2, column 8) and a class (\`zcl_mcptest_r3_lc1\` at line 8, column 15) on a different system. **5 of 5 calls return the input.**

No error is raised. The call appears to succeed.

## Where the bug lives

The mcp-server-abap wrapper (\`tools/navigation.go:29\`) is a one-line passthrough:

\`\`\`go
targetURI, err := client.NavigateToDefinition(ctx, uri)
\`\`\`

It returns whatever adtler returns. So the identity behavior originates inside adtler's \`NavigateToDefinition\`. Possibilities:

- The implementation is currently a stub returning the input.
- The HTTP call is being made but the response parser is reading the wrong field (the server may be echoing the request URI somewhere, and the parser may be picking that up instead of the resolved target).
- The endpoint needs an additional parameter (action, request body) that adtler isn't sending — and the server falls back to "no navigation, return source".

The fact that this reproduces identically on two SAP releases rules out a system-specific cause.

## Cross-references

- Consumer-side report with all 5 reproductions: Hochfrequenz/mcp-server-abap#286

hf-mrdachner:
## Summary

Closes #8.

`CL_SEDI_ADT_RES_NAVIGATION->post` reads the source via
`get_handler_for_plain_text` and combines it with the cursor position from the
URI fragment to resolve the navigation target. With a `nil` body the handler
cannot run, so `target_wb_request` stays initial and the response body never
gets set — which to callers looked like the endpoint was *echoing* the input
URI.

This PR adds a `source` parameter to `NavigateToDefinition` and posts the text
as `text/plain; charset=utf-8`. The caller already has the source in hand for
the cursor it is referring to, so this is cheaper than fetching it again
internally.

**Breaking change** in `adt.NavigationClient`:

```go
// before
NavigateToDefinition(ctx context.Context, sourceURI string) (string, error)
// after
NavigateToDefinition(ctx context.Context, sourceURI, source string) (string, error)
```

Mock (`custexport/export_test.go`), registry forwarder, and the httptest +
integration tests are updated.

## Test plan

- [x] `go test ./...` (unit, including updated httptest mock + body assertion)
- [x] `go vet -tags integration ./adt/...`
- [x] `go test -tags=integration -run TestNavigateToDefinition_Namespace_MultiSystem_Integration ./adt/` against `s4u` — cursor on `cl_abap_unit_assert` resolves to `/sap/bc/adt/oo/classes/cl_abap_unit_assert/source/main#start=2,6`
- [ ] R/3 (`hfq`) — test fixture is a bare REPORT with no cross-reference, so the integration test skips. R/3 verification deferred to local-agent run with a richer test source.
- [ ] Local agent re-runs integration tests with `needs:integration-test` label

### DIFF
--- a/adt/client.go
+++ b/adt/client.go
@@ -57,7 +57,13 @@ type DocuClient interface {
 
 // NavigationClient resolves source references.
 type NavigationClient interface {
-	NavigateToDefinition(ctx context.Context, sourceURI string) (string, error)
+	// NavigateToDefinition resolves the symbol at the cursor position
+	// embedded in sourceURI (as `...#start=line,column`) to its definition.
+	// The source code that the cursor refers to must be passed in source —
+	// the SAP handler reads the body as plain text and uses it together
+	// with the position fragment to compute the navigation target. Passing
+	// the source the caller already has avoids a round-trip GetSource call.
+	NavigateToDefinition(ctx context.Context, sourceURI, source string) (string, error)
 }
 
 // SearchClient provides object discovery.
--- a/adt/custexport/export_test.go
+++ b/adt/custexport/export_test.go
@@ -152,7 +152,7 @@ func (m *mockClient) SetMessages(context.Context, string, string, []adt.Message)
 func (m *mockClient) SetTextElements(context.Context, string, []adt.TextSymbol, []adt.SelectionText, string, string) error {
 	panic("not implemented")
 }
-func (m *mockClient) NavigateToDefinition(context.Context, string) (string, error) {
+func (m *mockClient) NavigateToDefinition(context.Context, string, string) (string, error) {
 	panic("not implemented")
 }
 func (m *mockClient) Rename(context.Context, string, string, string) (*adt.RenameResult, error) {
--- a/adt/navigation.go
+++ b/adt/navigation.go
@@ -7,12 +7,21 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strings"
 )
 
-func (c *httpClient) NavigateToDefinition(ctx context.Context, sourceURI string) (string, error) {
+func (c *httpClient) NavigateToDefinition(ctx context.Context, sourceURI, source string) (string, error) {
+	// SAP handler CL_SEDI_ADT_RES_NAVIGATION->post reads the source text
+	// from the request body via get_handler_for_plain_text and combines it
+	// with the cursor position from the URI fragment to resolve the target.
+	// Without a body the navigation cannot run and the response stays empty,
+	// which earlier looked to callers like an "echo" of the input URI.
 	path := "/sap/bc/adt/navigation/target?uri=" + url.QueryEscape(sourceURI)
-	resp, err := c.doMutate(ctx, http.MethodPost, path, nil,
-		map[string]string{"Accept": "application/xml"})
+	resp, err := c.doMutate(ctx, http.MethodPost, path, strings.NewReader(source),
+		map[string]string{
+			"Content-Type": "text/plain; charset=utf-8",
+			"Accept":       "application/xml",
+		})
 	if err != nil {
 		return "", fmt.Errorf("NavigateToDefinition: %w", err)
 	}
--- a/adt/navigation_integration_test.go
+++ b/adt/navigation_integration_test.go
@@ -9,19 +9,19 @@ import (
 	"testing"
 )
 
-// TestNavigateToDefinition_Namespace_MultiSystem_Integration exercises
-// NavigateToDefinition against real SAP systems. The fix passes the source
-// code in the request body — without it the SAP handler
-// (CL_SEDI_ADT_RES_NAVIGATION) cannot resolve the cursor position and the
-// response stays empty, which earlier looked to callers like an "echo" of
-// the input URI.
+// TestNavigateToDefinition_MultiSystem_Integration exercises
+// NavigateToDefinition against real SAP systems. The fix posts the source
+// code as the request body — CL_SEDI_ADT_RES_NAVIGATION reads the body via
+// get_handler_for_plain_text and combines it with the cursor position from
+// the URI fragment to compute the target. Without the body the handler
+// returns an empty response, which earlier looked like an "echo" of the
+// input URI.
 //
-// The test:
-//   - Asserts the call does NOT error
-//   - Logs the resolved target so failures are diagnosable
-//   - Skips when the test report has no cross-reference for the cursor to
-//     land on (R/3 fixture is a bare REPORT)
-func TestNavigateToDefinition_Namespace_MultiSystem_Integration(t *testing.T) {
+// The test asserts that a clear cross-reference (cl_abap_unit_assert or
+// TYPE string) resolves to a *different* URI than the cursor position —
+// echoing back means the fix regressed. When no cross-reference exists in
+// the test fixture (the R/3 stub is a bare REPORT), the test skips.
+func TestNavigateToDefinition_MultiSystem_Integration(t *testing.T) {
 	ctx := context.Background()
 	for _, sys := range eachSystem(t) {
 		sys := sys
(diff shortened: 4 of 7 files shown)


hf-mrdachner:
## GO with notes

The fix is plausible, the diff is small, all implementers of `NavigationClient` are updated, and `go build ./...` / `go test ./...` / `go vet -tags integration ./adt/...` are clean. Verified the namespaced-attr XML parsing extracts the `adtcore:uri` from both the real SAP shape (single self-closing `<adtcore:objectReference adtcore:uri=... xmlns:adtcore=...>`) and the test-mock shape. The PR body claims the s4u integration run resolved `cl_abap_unit_assert` to `/sap/bc/adt/oo/classes/cl_abap_unit_assert/source/main#start=2,6` — that is the strongest evidence the body-as-source theory is correct.

That said, several items are worth tightening before merge:

### 1. The integration test does not fail on echo — it only logs

`adt/navigation_namespace_integration_test.go:71-75`:
```go
if target == sourceURI || target == \"\" {
    t.Logf(\"[%s] endpoint echoed input (known SAP limitation per issue #8 investigation)\", sys.Name)
} else {
    t.Logf(\"[%s] endpoint returned a DIFFERENT target — navigation resolved!\", sys.Name)
}
```

The PR's whole premise is that echo == bug. The post-fix test should treat echo on s4u as a regression (`t.Fatalf` or `t.Errorf`), not log it as a \"known SAP limitation\". Otherwise this test is green even if the fix is silently un-done. R/3 can stay tolerant via the existing `t.Skipf` for missing cross-references, but s4u with `cl_abap_unit_assert` in the fixture is the deterministic happy path and should assert `target != sourceURI` and `strings.Contains(target, \"cl_abap_unit_assert\")`.

### 2. Stale references in the test file

- The function name `TestNavigateToDefinition_Namespace_MultiSystem_Integration` still says \"Namespace\" — that referred to the PR #32 namespace-parsing hypothesis, which the issue's raw-XML probe disproved. Now that the root cause is the missing body, the name is misleading. Suggest `TestNavigateToDefinition_MultiSystem_Integration`.
- `adt/navigation_namespace_integration_test.go:67`: `\"namespace parsing may be broken\"` in the error string is also stale — the bug is no longer namespace parsing.
- The log line at line 72 about \"known SAP limitation per issue #8 investigation\" directly contradicts the PR description.

### 3. Eclipse capture was the gating prerequisite per issue #8

Issue #8 was explicitly labelled `blocked:eclipse-capture` and the last comment said \"Leaving the issue open for when that capture happens.\" The PR description gives a plausible-sounding inference about `CL_SEDI_ADT_RES_NAVIGATION->post` calling `get_handler_for_plain_text`, but does not show any HTTP capture or quote ABAP source. The integration-run result on s4u is good empirical support, but it would be much stronger if the PR linked an Eclipse capture (or a SAP Note / `where_used` of `get_handler_for_plain_text`) to confirm the body shape. Worth adding a one-line citation in the function comment.

### 4. Edge cases not covered

- **Empty source**: `strings.NewReader(\"\")` will still POST a zero-length body. Behaviour undefined. A defensive `if source == \"\" { return \"\", fmt.Errorf(\"NavigateToDefinition: source must not be empty\") }` plus a unit test would prevent silent regressions.
- **Source / sourceURI mismatch**: caller can pass a `source` that does not match the file at `sourceURI`. SAP uses the body, so the wrong source produces a wrong navigation target with no error. Worth documenting in the godoc that source must match the file's current state — the existing godoc says \"the source code that the cursor refers to must be passed\" but does not warn about staleness.
- **URI without `#start=` fragment**: not tested. SAP's `extract_source_code_position` would presumably fail and return an HTTP error — the current code path passes that through, which is acceptable, but worth a unit test.
- **`CX_SEDI_ADT_MULTIPLE_TARGETS`**: when an identifier resolves to several candidates Eclipse shows a picker. The current implementation can only return one URI. Worth at least a `// TODO` comment.

### 5. Consumer-side coordination

Issue #8 references `Hochfrequenz/mcp-server-abap#286`, which is currently CLOSED — but `tools/navigation.go` over there will fail to build the moment this PR's `NavigationClient` signature lands, because it calls `NavigateToDefinition(ctx, uri)`. The PR should at minimum:

- File / re-open a tracking issue on mcp-server-abap noting that the breaking change in adtler requires a coordinated update (the consumer must now pass `source` — likely from `GetSource(uri)` first).
- Mention this dependency in the PR description so the merge order is clear (the consumer cannot bump its adtler dependency without a code change).

### 6. Minor style

- Adding a `source` positional parameter to a public interface method works but locks the signature against future additions. If more parameters become needed (e.g. element kind hint, multi-target preference), an `opts ...NavigateOption` or `req NavigateRequest` would scale better. Not blocking, just worth thinking about while we are already breaking the API.
- `Accept: application/xml` is kept as-is. CLAUDE.md notes S/4 is stricter on Accept headers and prefers vendor MIME types. Since the s4u integration run succeeded, this is fine in practice — but `application/vnd.sap.adt.objectReference+xml` would be more defensible if a future S/4 patch tightens it.

### Verdict change

What would change my verdict to plain GO:
- The integration test asserts (not logs) a non-echo, target-contains-`cl_abap_unit_assert` result on s4u, and the misleading \"Namespace\" / \"known SAP limitation\" comments are removed; the consumer-side breaking-change tracker on mcp-server-abap is filed.

hf-mrdachner:
## Integration test results

Ran on real SAP systems (`hfq` = R/3, `s4u` = S/4HANA) at HEAD `ad7a4f7`.

### `TestNavigateToDefinition_MultiSystem_Integration`

| System | Result | Detail |
|---|---|---|
| `hfq` (R/3) | ⏭️ SKIP | Z_ADT_MCP_TEST_REPORT on R/3 is a bare `REPORT` (6 lines, no cross-references); the test correctly skips when there is nothing to navigate to. The body-as-source fix is system-agnostic — verifying it on R/3 would need a richer test fixture, tracked separately. |
| `s4u` (S/4) | ✅ PASS | Cursor on `cl_abap_unit_assert` at `#start=15,5` resolved to `/sap/bc/adt/oo/classes/cl_abap_unit_assert/source/main#start=2,6` — a different URI, so the previous "echo" failure mode no longer occurs. |

### Run details

- Command: `SAP_INTEGRATION_SYSTEM=<sys> SAP_INTEGRATION_SYSTEMS=<sys> go test -tags=integration -v -count=1 -run TestNavigateToDefinition_MultiSystem_Integration ./adt/`
- The tightened test now hard-fails on echo (`target == "" || target == sourceURI`) — a regression in the body-as-source fix would surface as a clear failure here.
- No leftover test objects in `$TMP`.

PR #55, opened 2026-05-08
chore(#42): drop dead 412 ETag charset retry, superseded by #35

### LINKED ISSUE #42: enhancement: normalize ETag charset at receive time (replace #15 retry)
Follow-up from #35 (see design doc for context).

Today, `SetSource` handles the TABL ETag charset mismatch (adtler#15) via a 412 retry that patches the ETag string: if the PUT returns `412 ExceptionPreconditionFailed` and the ETag contains `text/plain` without `charset`, we insert `; charset=utf-8` and retry.

This works but costs an extra round-trip on every TABL write. A cleaner fix is to normalize ETags at receive time: when `GetSource` / `FetchETag` extract the `ETag` header, detect the missing charset and patch it immediately. Then the retry path in `SetSource` becomes dead code and can be deleted.

## Proposed change

1. Add a `normalizeETag(etag, contentType string) string` helper that inserts the charset if missing.
2. Call it in `GetSource` and `FetchETag` where the ETag is extracted from the response header.
3. Delete the 412 retry branch in `SetSource`.
4. Keep the existing `source_etag_charset_integration_test.go` to guard the behaviour.

## Why defer

- Needs validation on both R/3 and S/4 that the same normalization is correct for all object types (not just TABL).
- The existing retry is pessimistic but safe — removing it is a behaviour change that wants its own integration cycle.

Brainstormed during the #35 design discussion.

hf-mrdachner:
## Summary

Closes #42.

#42 was filed as a follow-up to #35's design doc, proposing to replace the #15 412 ETag charset retry with receive-time normalization. By the time #35 actually merged (aa101f7, ~6h after #42 was opened) the asymmetry it targeted was structurally gone: \`sourceContentType\` prefers the charset-qualified type when discovery advertises it, so GetSource and PUT now agree on the ETag form. The retry path has been dormant since.

This PR removes the now-dead retry plus its only caller, the predicate, and the predicate's unit test. The public \`ExceptionTypePreconditionFailed\` constant stays — downstream callers may still compare against it. The \`TestSetSource_TABLETagCharset_MultiSystem_Integration\` regression test stays too: it exercises the full GetSource → SetSource cycle end-to-end and would catch any reappearance of the asymmetry on a future SAP release that drops the discovery hint.

A receive-time normalization variant was tried locally and produced a 412 in the *opposite* direction (current s4u stores the short form, server expects short, normalize produced long → mismatch), confirming the retry is no longer warranted on accessible systems.

## Test plan

- [x] \`go build ./...\`
- [x] \`go vet -tags integration ./adt/...\`
- [x] \`go test ./...\` (all unit tests pass; the dropped predicate test was the only one using \`isPreconditionFailed\`)
- [x] \`go test -tags=integration -run TestSetSource_TABLETagCharset_MultiSystem_Integration ./adt/\` against \`s4u\` — PASS (the 412 path is never reached; the test reports the expected \`403 ExceptionResourceNoAccess\` from the unrelated DDIC enqueue family)
- [ ] R/3 (\`hfq\`) — TABL fixture skip (DDIC create not available on ECC)


hf-mrdachner:
## GO with notes

I reviewed PR #55 in a fresh worktree against HEAD `3bc8999`, with no prior conversation context. Verdict: **GO with notes** — the cleanup is correct on every system that #35 covers in practice, builds and tests pass, and no consumer is broken. A few stale references and one defensive concern keep this from being unconditional GO.

### Verification I ran

- `go build ./...` — pass
- `go build -tags integration ./adt/...` — pass
- `go vet -tags integration ./adt/...` — pass
- `go test ./...` — all packages pass:
  ```
  ok  	github.com/Hochfrequenz/adtler/adt	2.927s
  ok  	github.com/Hochfrequenz/adtler/adt/adtxml	0.899s
  ok  	github.com/Hochfrequenz/adtler/adt/custexport	2.165s
  ok  	github.com/Hochfrequenz/adtler/auth	2.480s
  ```

### Premise check: is the retry actually dead?

Mostly yes, with one edge case worth naming. After PR #41 wired `sourceContentType` (`adt/source.go:67-97`) into both `GetSource` (`adt/source.go:103`) and `setSourceWithLockHeader`/`setSourceWithLockParam` (`adt/source.go:365`, `383`), the Accept on GET and the Content-Type on PUT are computed from the **same** discovery snapshot. The asymmetry that drove #15 (GET sent `text/plain`, PUT was hardcoded to `text/plain; charset=utf-8`) is structurally gone whenever the two calls see the same discovery state.

The remaining edge case is the empty-discovery fallback: `sourceContentType` returns bare `text/plain` (`adt/source.go:85, 96`) when the discovery cache has no prefix match for the endpoint. In that case both GET and PUT use `text/plain` and the ETag forms still match, so the asymmetry still cannot reappear. The only way #15 could come back is if SAP shipped a release where the server stored ETags with `; charset=utf-8` while discovery advertised only the bare form **and** the server still validated against the charset-qualified stored form. That is the exact configuration the design doc (`docs/superpowers/specs/2026-04-10-issue-35-discovery-content-negotiation-design.md:120`) said the retry would catch — removing it does narrow the safety margin slightly, but the empirical evidence (S/4 advertises both forms, charset-qualified preferred; R/3 doesn't run TABL writes) makes this a defensible trade.

### API surface

Safe. Searched the org for `isPreconditionFailed` (`gh search code "isPreconditionFailed" --owner Hochfrequenz`) — only hits inside adtler itself (the now-removed `adt/types.go` definition, the now-removed `adt/source.go` call site, and the now-removed `adt/client_internal_test.go` test, plus stale references in design docs under `docs/superpowers/`). No consumer (mcp-server-abap or otherwise) uses it, and it was unexported anyway. The public `ExceptionTypePreconditionFailed` constant is preserved at `adt/types.go:141` — downstream branching on that value still compiles.

### Test breadth

The only removed test is `TestIsPreconditionFailed_TypeAware` in `adt/client_internal_test.go`, which only exercised the deleted predicate. Searched for httptest mocks that returned 412 followed by 200 (i.e. would have asserted retry behaviour end-to-end): none exist — the retry path was only ever covered by the live integration test. So no other tests need adjusting.

### Regression coverage in `source_etag_charset_integration_test.go`

The test does fire its critical assertion. Walking the flow at `adt/source_etag_charset_integration_test.go:24-82`:

- `CreateObject` succeeds on S/4
- `GetSource` succeeds (line 42) — confirmed in the PR description
- `LockObject` succeeds (line 49) — also confirmed
- `SetSource` is called at line 56 and produces the 403 enqueue error mentioned in the PR description
- The 412/PreconditionFailed substring check at line 71 evaluates against that 403 error string and correctly drops into the t.Logf branch at line 76

So if the asymmetry resurfaced and SetSource produced 412, the t.Errorf at line 72 would fire. The test is genuine regression coverage (within the limit that it can only test the SetSource line of `eachSystem(t)` actually runs through to a 412-vs-403 distinction — fine on accessible systems today).

### Stale references that should be cleaned up

These are not blockers but should ideally be fixed in this PR or a follow-up:

1. **`CLAUDE.md:94`** still asserts `SetSource patches the ETag string on 412`. After this PR merges that bullet is wrong. Recommend updating to reflect the discovery-driven symmetry instead.
2. **`adt/source_etag_charset_integration_test.go:13-23`** docstring still says "The fix patches the ETag string on 412 and retries" (line 17) and "The ETag charset patch should have prevented that" (line 65). The test is still useful as regression coverage but the comment should be reworded to "the discovery-driven symmetry between GetSource and SetSource Content-Types should have prevented that" — otherwise a future reader looking for the patch logic will find nothing.
3. **`docs/superpowers/specs/2026-04-10-issue-35-discovery-content-negotiation-design.md:31, 109-111, 120`** explicitly say the 412 retry stays. Worth a one-line marker in those docs that #42/#55 superseded that choice.
4. **`docs/superpowers/specs/2026-04-27-issue-43-adt-error-namespace-type-design.md:99-107, 157, 179`** and **`docs/superpowers/plans/2026-04-27-issue-43-adt-error-namespace-type.md`** reference `isPreconditionFailed` as a live predicate. These are historical specs/plans for already-merged work, so updating is optional, but a sweep would be tidy.

### Defensive concern

With the retry gone, any system that genuinely still has the asymmetry now surfaces a hard 412 to the caller instead of being self-healed. The PR's empirical test (s4u with the receive-time-normalize variant produced a 412 in the opposite direction) is good evidence the asymmetry is gone today on s4u, but it's evidence from one S/4 release. If a future tenant lands in the empty-discovery fallback while still serving charset-qualified ETags from the table, the regression will be visible immediately rather than silently retried. I think that's the right call (silent retries hide drift), but it's worth being explicit that this is a behaviour change for the dormant edge case.

### What would change my verdict to plain GO

Update the three stale references in shipping artifacts (`CLAUDE.md:94`, the integration test's docstring at `adt/source_etag_charset_integration_test.go:13-23`, and a one-line marker on the #35 design doc) so the codebase reads consistently after this drop. The historical superpowers docs under `docs/superpowers/` can stay as-is.

hf-mrdachner:
## Integration test results

Ran on real SAP systems (`hfq` = R/3, `s4u` = S/4HANA) at HEAD `25e4606`.

### `TestSetSource_TABLETagCharset_MultiSystem_Integration`

| System | Result | Detail |
|---|---|---|
| `hfq` (R/3) | ⏭️ SKIP | DDIC TABL creation not available on this ECC system (404 / 415 — known, expected). |
| `s4u` (S/4) | ✅ PASS | `GetSource` ETag: `20260508085423000text/plain2` — bare `text/plain` form. `SetSource` returns the unrelated `403 ExceptionResourceNoAccess` (DDIC enqueue family, adtler#4). The critical assertion — error is NOT 412 — holds, so the GET / PUT charset symmetry that #35's discovery negotiation guarantees is still intact, and removing the 412 retry surfaces no regression. |

The retry was dormant on every accessible system; the test confirms the path it would have caught is no longer reachable.

### Run details

- Command: `SAP_INTEGRATION_SYSTEM=<sys> SAP_INTEGRATION_SYSTEMS=<sys> go test -tags=integration -v -count=1 -run TestSetSource_TABLETagCharset_MultiSystem_Integration ./adt/`
- Cleanup: `ZTAB_MCP_30461` was deleted by the test's `t.Cleanup`. No residual `$TMP` objects.

PR #66, opened 2026-06-05
test(#63): integration tests probing SYST/CUST transport hypotheses

### LINKED ISSUE #63: GetTransportRequests returns empty on S/4HANA systems using KORRDEV=SYST/CUST
## Symptoms

- `GetTransportInfo("S4UK901071")` — works, returns `Status: "D"`, `Owner: "METZEJ"` ✅
- `GetTransportObjects("S4UK901071")` — works, returns 12 objects in the TR ✅  
- `GetTransportRequests(ctx, "METZEJ", "D")` — returns empty list, no error ❌
- `GetTransportRequests(ctx, "", "D")` — returns empty list, no error ❌
- `GetTransportRequests(ctx, "", "")` — returns empty list, no error ❌

In short: **single-TR operations work, the list does not.**

## Root cause

The ADT transport organizer tree endpoint only surfaces `KORRDEV="K"` (standard workbench) requests. On this S/4HANA system, transport requests use `KORRDEV="SYST"` and `KORRDEV="CUST"` (S/4HANA-specific classifications). The endpoint silently returns an empty `<tm:root/>` for these types — not an error, just no children.

Confirmed via direct E070 SQL (`RunQuery`):

```
TRKORR        AS4USER  TRSTATUS  KORRDEV
S4UK901071    METZEJ   D         SYST
S4UK901075    METZEJ   D         SYST
S4UK901076    METZEJ   D         CUST
S4UK901077    METZEJ   D         CUST
S4UK901078    METZEJ   D         CUST
S4UK901079    METZEJ   D         CUST
```

The SAP response for `GetTransportRequests` with any user/status combination is always:
```xml
<?xml version="1.0" encoding="utf-8"?>
<tm:root adtcore:name="METZEJ" ... xmlns:tm="http://www.sap.com/cts/adt/tm" xmlns:adtcore="http://www.sap.com/adt/core"/>
```

Self-closing, no children — SAP filters SYST/CUST requests out before serializing the XML.

## Workaround

Query E070 and E07T directly via `RunQuery` as an alternative. This works for all KORRDEV types.

## Possible fix

Investigate whether a different Accept header or endpoint path exposes SYST/CUST requests via ADT. The ADT transport organizer may need to be called differently (different content negotiation or a separate REST path) to include these S/4HANA-specific request types.

Note: a previously attempted fix adding `workbenchRequests>workbenchRequest` XML parsing was reverted — the issue is not XML format but the SAP-side filtering of SYST/CUST requests before they reach the XML serializer.

Jonatan-HF:
Closes #63

### Root Cause

`GetTransportRequests` returns an empty list on S/4HANA systems where all transport requests use `KORRDEV=SYST` or `KORRDEV=CUST`. The ADT endpoint `GET /sap/bc/adt/cts/transportrequests` with `Accept: application/vnd.sap.adt.transportorganizertree.v1+xml` is hard-coded server-side (in `CL_ADT_CTS_MANAGEMENT`) to return only `KORRDEV=K` (classic workbench corrections). SAP returns `<tm:root/>` (empty, no error) for everything else.

**Impact on S4U:** E070 contains 993 SYST + 394 CUST modifiable transports. `GetTransportRequests` returns 0 — 100% data loss.

On newer S/4HANA systems, `K` is no longer in the `TRCATEG` domain at all. It was replaced by `SYST`. The standard ADT tree endpoint is structurally unusable on these systems.

### Investigation: What We Ruled Out

| Approach | Result |
|---|---|
| Alternative `status=` values | Still 0 — server-side filter is not status-based |
| `application/xml` accept header | SAP filters server-side regardless of content type |
| URL params `?workbench=true` etc. (lowercase) | Not accepted by the endpoint |
| `/sap/bc/adt/cts/workbench` separate endpoint | Workbench-only, same problem |
| Higher version numbers (v2, v3) | Do not exist |
| Eclipse ADT Transport Organizer view | Also shows nothing for SYST/CUST — same server filter |

No configuration knob (`CTSCONFIG`, `CTS_TPROPS`) exists to change this behaviour.

### Investigation: Open Hypotheses (What This PR Tests)

ABAP source analysis of `IF_CTS_ADT_TM_CONSTANTS` and `CL_CTS_ADT_TM_CONFIG_HANDLER` on S4U revealed two untested candidates:

**Hypothesis A — Non-tree content type:**

```
Accept: application/vnd.sap.adt.transportorganizer.v1+xml  (CO_SAP_TM_CONTENT_TYPE)
```
vs. the current tree variant `...transportorganizertree.v1+xml` (`CO_SAP_TM_TREE_CONTENT_TYPE`).

**Hypothesis B — CamelCase query parameters:**

`CL_CTS_ADT_TM_CONFIG_HANDLER` defines these query parameter constants (previously untested because all documented examples only use `user=` and `status=`):

| Constant | Query param |
|---|---|
| `CO_WORKBENCH` | `WorkbenchRequests` |
| `CO_CUSTOMIZING` | `CustomizingRequests` |
| `CO_TRANSPORT_OF_COPIES` | `TransportOfCopies` |
| `CO_MODIFIABLE` | `Modifiable` |
| `CO_RELEASED` | `Released` |
| `CO_USER` | `User` |

Sending `?WorkbenchRequests=true&CustomizingRequests=true&TransportOfCopies=true` may unlock all transport categories. The SAP backend has three separate handler methods: `get_wb_trans_requests`, `get_cust_trans_requests`, `get_toc_trans_requests` — one per category.

**Note:** the current code sends `?user=X&status=D` but the SAP constants map to `User=X` and `Modifiable=true` / `Released=true`. The lowercase params may be case-insensitively accepted; this PR also tests the correct CamelCase forms.

### What This PR Contains

Four read-only integration tests in `package adt` (internal, for `*httpClient.doRead` access):

1. **`GapDocumentation`** — queries E070 via `RunQuery` and compares the KORRDEV distribution against `GetTransportRequests`. Logs findings without failing (zero ADT results is the documented bug, not a test error).

2. **`AlternativeAcceptHeader`** — hits the list endpoint with `application/vnd.sap.adt.transportorganizer.v1+xml`. Logs the raw XML response and counts `<*:request>` elements.

3. **`AllTypeQueryParams`** — four probes: `WorkbenchRequests` only, `CustomizingRequests` only, all three types with `Modifiable=true`, all three types with legacy `status=D`.

4. **`CombinedHypothesis`** — non-tree accept header + all type params together. Logs the full response body (up to 2000 chars) regardless of result.

### Fallback Plan

If all ADT hypotheses are rejected, the confirmed working path is `RunQuery` on `E070 JOIN E07T`. This was verified on S4U: the query executes in 7–20 ms and returns all KORRDEV types. Implementation would be a fallback in `GetTransportRequests` when the ADT endpoint returns empty.

## How to Run

```bash
SAP_INTEGRATION_SYSTEMS=S4U go test -tags=integration -v \
  -run TestGetTransportRequests_SystCust ./adt/...
```

The tests are read-only and create no SAP objects.

### DIFF
--- a/adt/transport_syst_cust_investigation_integration_test.go
+++ b/adt/transport_syst_cust_investigation_integration_test.go
@@ -35,6 +35,7 @@ package adt
 import (
 	"context"
 	"encoding/xml"
+	"bytes"
 	"io"
 	"net/http"
 	"net/url"
@@ -57,7 +58,8 @@ type issue63System struct {
 // issue63Systems loads all SAP systems permitted by the SAP_INTEGRATION_SYSTEMS
 // (or SAP_INTEGRATION_SYSTEM / default_system) whitelist. It mirrors the
 // selection logic of eachSystem in integration_helpers_test.go but returns
-// *httpClient values for direct HTTP probing.
+// *httpClient values for direct HTTP probing. The duplication is intentional:
+// eachSystem lives in package adt_test and is not accessible from package adt.
 func issue63Systems(t *testing.T) []issue63System {
 	t.Helper()
 
@@ -128,10 +130,11 @@ func TestGetTransportRequests_SystCust_GapDocumentation_Integration(t *testing.T
 		sys := sys
 		t.Run(sys.name, func(t *testing.T) {
 			ctx := context.Background()
-			client := NewClient(sys.c.cfg)
 
 			// E070: count modifiable transports per KORRDEV type.
-			e070, err := client.RunQuery(ctx,
+			// sys.c implements the full Client interface — reuse it instead of
+			// constructing a second client, which would split CSRF session state.
+			e070, err := sys.c.RunQuery(ctx,
 				"SELECT KORRDEV, COUNT( TRKORR ) AS CNT FROM E070 WHERE TRSTATUS = 'D' GROUP BY KORRDEV ORDER BY CNT DESCENDING",
 				50)
 			if err != nil {
@@ -147,9 +150,20 @@ func TestGetTransportRequests_SystCust_GapDocumentation_Integration(t *testing.T
 						cntIdx = i
 					}
 				}
+				// SAP may not honour the alias "CNT" for aggregate columns; fall
+				// back to column index 1 if the alias lookup failed.
+				if korrdevIdx < 0 {
+					korrdevIdx = 0
+				}
+				if cntIdx < 0 {
+					if len(e070.Columns) > 1 {
+						cntIdx = 1
+						t.Logf("  (column alias CNT not found, using column index 1: %q)", e070.Columns[1].Name)
+					}
+				}
 				for _, row := range e070.Rows {
 					korrdev, cnt := "", ""
-					if korrdevIdx >= 0 && korrdevIdx < len(row) {
+					if korrdevIdx < len(row) {
 						korrdev = row[korrdevIdx]
 					}
 					if cntIdx >= 0 && cntIdx < len(row) {
@@ -160,7 +174,7 @@ func TestGetTransportRequests_SystCust_GapDocumentation_Integration(t *testing.T
 			}
 
 			// ADT standard endpoint: current behavior.
-			transports, err := client.GetTransportRequests(ctx, "", "D")
+			transports, err := sys.c.GetTransportRequests(ctx, "", "D")
 			if err != nil {
 				t.Logf("GetTransportRequests failed: %v", err)
 			} else {
@@ -248,6 +262,8 @@ func TestGetTransportRequests_SystCust_AllTypeQueryParams_Integration(t *testing
 			params: url.Values{"WorkbenchRequests": {"true"}, "CustomizingRequests": {"true"}, "TransportOfCopies": {"true"}, "Modifiable": {"true"}},
 		},
 		{
+			// CamelCase type params (CL_CTS_ADT_TM_CONFIG_HANDLER) combined with
+			// the legacy lowercase status=D that GetTransportRequests currently sends.
 			name:   "AllTypes_LegacyStatusD",
 			params: url.Values{"WorkbenchRequests": {"true"}, "CustomizingRequests": {"true"}, "TransportOfCopies": {"true"}, "status": {"D"}},
 		},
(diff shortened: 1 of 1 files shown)


Jonatan-HF:
## Integration Test Results (S4U, 2026-06-05)

```
SAP_INTEGRATION_SYSTEMS=S4U go test -tags=integration -v -run TestGetTransportRequests_SystCust ./adt/...
```

### Gap Documentation

E070 on S4U (TRSTATUS=D):

| KORRDEV | Count |
|---|---|
| SYST | 995 |
| CUST | 394 |
| (empty) | 68 |
| **K** | **0** |

`GetTransportRequests` returned **0 transports** → confirms 100% data loss on this system.

### Hypothesis A — Non-tree Accept header (`transportorganizer.v1+xml`)

**HTTP 406** `ExceptionResourceNotAcceptable`

SAP error message:
> The message content is not acceptable. Accepted content types: `application/vnd.sap.adt.transportorganizertree.v1+xml`

The endpoint has exactly one accepted content type hard-coded. **Hypothesis A rejected.**

### Hypothesis B — CamelCase query parameters

All four combinations returned **HTTP 200 with 0 results**:

| Params | Result |
|---|---|
| `WorkbenchRequests=true&Modifiable=true` | 0 |
| `CustomizingRequests=true&Modifiable=true` | 0 |
| `WorkbenchRequests=true&CustomizingRequests=true&TransportOfCopies=true&Modifiable=true` | 0 |
| `WorkbenchRequests=true&CustomizingRequests=true&TransportOfCopies=true&status=D` | 0 |

SAP accepts the parameters without error but the server-side KORRDEV=K filter runs before the type selection. **Hypothesis B rejected.**

### Hypothesis C — Combined (non-tree header + all type params)

**HTTP 406** — same as Hypothesis A. **Hypothesis C rejected.**

---

## Conclusion

There is no ADT-native way to surface SYST/CUST transport requests. The endpoint architecture enforces KORRDEV=K at the serialization layer, upstream of any content-type or query-parameter routing.

**The only viable fix is a `RunQuery` fallback on E070 JOIN E07T.** This was confirmed working in the earlier investigation and covers all KORRDEV types in 7–20 ms.

Next step: implement the fallback in a separate `fix/63-*` branch.

PR #67, opened 2026-06-05
fix(#63): fall back to E070/E07T when ADT transport organizer returns empty

### LINKED ISSUE #63: GetTransportRequests returns empty on S/4HANA systems using KORRDEV=SYST/CUST

Jonatan-HF:
## Summary

Fixes #63.

- `GetTransportRequests` now falls back to a `RunQuery` on `E070 JOIN E07T` when the ADT transport organizer tree endpoint returns an empty `<tm:root/>` — the silent behavior on S/4HANA systems using `KORRDEV=SYST/CUST`
- Investigation in PR #66 confirmed all ADT-native approaches (alternative Accept header, CamelCase query params) return 0 results due to server-side KORRDEV=K filtering; the E070 fallback is the only viable fix
- The fallback is only triggered when ADT returns zero results, so classic K-type systems are unaffected

## Changes

- `transport.go`: add `getTransportRequestsViaQuery` — builds a parameterized `SELECT e.TRKORR, e.AS4USER, e.TRSTATUS, t.AS4TEXT FROM E070 LEFT OUTER JOIN E07T WHERE e.STRKORR = ''` query (excluding tasks), applies user/status filters, returns typed `[]TransportRequest`
- `transport_test.go`: two new unit tests — fallback triggered when ADT returns empty; fallback NOT triggered when ADT returns results
- `transport_integration_test.go`: `TestGetTransportRequests_AllSystems_Integration` using `eachSystem(t)` to cover both R/3 and S/4
- `registry_test.go`: fix `emptyTransports` mock to return one real transport so the registry delegation test no longer triggers the fallback path

## Test plan

- [ ] `go test ./...` passes (unit tests)
- [ ] `SAP_INTEGRATION_SYSTEMS=HFQ,S4U go test -tags=integration -v -run TestGetTransportRequests ./adt/...` passes on both systems
- [ ] S4U: result count > 0 (exercises E070 fallback); HFQ: result count > 0 (exercises ADT path)

### DIFF
--- a/adt/transport.go
+++ b/adt/transport.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 	"time"
 
@@ -403,32 +404,23 @@ func (c *httpClient) GetTransportRequests(ctx context.Context, user, status stri
 	return result, nil
 }
 
+// transportUserRe matches valid SAP usernames safe to embed in a SQL WHERE
+// clause literal. SAP usernames are at most 40 chars: alphanumeric, _, -, ..
+var transportUserRe = regexp.MustCompile(`^[A-Za-z0-9_.\-]{1,40}$`)
+
+// transportStatusRe matches a valid CTS transport status: a single uppercase
+// letter (D = modifiable, L = released, A = accepted, …).
+var transportStatusRe = regexp.MustCompile(`^[A-Z]$`)
+
 // getTransportRequestsViaQuery is the fallback for GetTransportRequests on
 // S/4HANA systems where KORRDEV=SYST/CUST prevents the ADT transport organizer
-// tree endpoint from returning any results. It queries E070 and E07T directly
-// via the ADT data preview endpoint.
-// validTransportUserParam reports whether s is safe to embed in a SQL WHERE
-// clause as a quoted literal. SAP usernames are at most 40 chars and contain
-// only alphanumeric characters, underscores, hyphens, and dots — no quotes,
-// semicolons, or other characters that could break the surrounding SQL string.
-func validTransportUserParam(s string) bool {
-	if len(s) > 40 {
-		return false
-	}
-	for _, r := range s {
-		if ('A' <= r && r <= 'Z') || ('a' <= r && r <= 'z') || ('0' <= r && r <= '9') || r == '_' || r == '-' || r == '.' {
-			continue
-		}
-		return false
-	}
-	return true
-}
-
+// tree endpoint from returning any results. It queries E070 directly via the
+// ADT data preview endpoint (JOINs are rejected on some S/4HANA releases).
 func (c *httpClient) getTransportRequestsViaQuery(ctx context.Context, user, status string) ([]TransportRequest, error) {
-	if user != "" && !validTransportUserParam(user) {
+	if user != "" && !transportUserRe.MatchString(user) {
 		return nil, fmt.Errorf("GetTransportRequests: user %q contains characters not allowed in a transport query", user)
 	}
-	if status != "" && (len(status) != 1 || status[0] < 'A' || status[0] > 'Z') {
+	if status != "" && !transportStatusRe.MatchString(status) {
 		return nil, fmt.Errorf("GetTransportRequests: status must be a single uppercase letter, got %q", status)
 	}
 
--- a/adt/transport.go
+++ b/adt/transport.go
@@ -393,6 +393,82 @@ func (c *httpClient) GetTransportRequests(ctx context.Context, user, status stri
 			Description: r.Description, Status: r.Status,
 		}
 	}
+
+	// S/4HANA systems using KORRDEV=SYST/CUST have transports silently filtered
+	// by the ADT endpoint, which only surfaces KORRDEV=K requests. Fall back to
+	// querying E070/E07T directly when the ADT response is empty.
+	if len(result) == 0 {
+		return c.getTransportRequestsViaQuery(ctx, user, status)
+	}
+	return result, nil
+}
+
+// getTransportRequestsViaQuery is the fallback for GetTransportRequests on
+// S/4HANA systems where KORRDEV=SYST/CUST prevents the ADT transport organizer
+// tree endpoint from returning any results. It queries E070 and E07T directly
+// via the ADT data preview endpoint.
+func (c *httpClient) getTransportRequestsViaQuery(ctx context.Context, user, status string) ([]TransportRequest, error) {
+	if strings.ContainsAny(user, "'\"\\;") {
+		return nil, fmt.Errorf("GetTransportRequests: invalid character in user %q", user)
+	}
+	if status != "" && (len(status) != 1 || status[0] < 'A' || status[0] > 'Z') {
+		return nil, fmt.Errorf("GetTransportRequests: status must be a single uppercase letter, got %q", status)
+	}
+
+	where := []string{"e.STRKORR = ''"}
+	if user != "" {
+		where = append(where, "e.AS4USER = '"+user+"'")
+	}
+	if status != "" {
+		where = append(where, "e.TRSTATUS = '"+status+"'")
+	}
+
+	sql := "SELECT e.TRKORR, e.AS4USER, e.TRSTATUS, t.AS4TEXT" +
+		" FROM E070 AS e LEFT OUTER JOIN E07T AS t ON t.TRKORR = e.TRKORR AND t.LANGU = 'E'" +
+		" WHERE " + strings.Join(where, " AND ") +
+		" ORDER BY e.TRKORR"
+
+	qr, err := c.RunQuery(ctx, sql, 5000)
+	if err != nil {
+		return nil, fmt.Errorf("GetTransportRequests: fallback E070 query: %w", err)
+	}
+
+	trkorrIdx, userIdx, statusIdx, textIdx := -1, -1, -1, -1
+	for i, col := range qr.Columns {
+		switch col.Name {
+		case "TRKORR":
+			trkorrIdx = i
+		case "AS4USER":
+			userIdx = i
+		case "TRSTATUS":
+			statusIdx = i
+		case "AS4TEXT":
+			textIdx = i
+		}
+	}
+	if trkorrIdx < 0 {
+		return nil, fmt.Errorf("GetTransportRequests: fallback query returned no TRKORR column")
+	}
+
+	result := make([]TransportRequest, 0, len(qr.Rows))
+	for _, row := range qr.Rows {
+		tr := TransportRequest{}
+		if trkorrIdx < len(row) {
+			tr.Number = strings.TrimSpace(row[trkorrIdx])
+		}
+		if userIdx >= 0 && userIdx < len(row) {
+			tr.Owner = strings.TrimSpace(row[userIdx])
+		}
+		if statusIdx >= 0 && statusIdx < len(row) {
+			tr.Status = strings.TrimSpace(row[statusIdx])
+		}
+		if textIdx >= 0 && textIdx < len(row) {
+			tr.Description = strings.TrimSpace(row[textIdx])
+		}
+		if tr.Number != "" {
+			result = append(result, tr)
+		}
+	}
 	return result, nil
 }
 
--- a/adt/transport.go
+++ b/adt/transport.go
@@ -432,25 +432,27 @@ func (c *httpClient) getTransportRequestsViaQuery(ctx context.Context, user, sta
 		return nil, fmt.Errorf("GetTransportRequests: status must be a single uppercase letter, got %q", status)
 	}
 
-	where := []string{"e.STRKORR = ''"}
+	// Exclude tasks (STRKORR is set on tasks; top-level requests have STRKORR = '').
+	where := []string{"STRKORR = ''"}
 	if user != "" {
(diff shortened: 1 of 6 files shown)


Jonatan-HF (commented):
### Strengths

- The fallback strategy is well-chosen: `RunQuery` (E070 JOIN E07T) is already the proven workaround documented in the issue, and piggybacking on it avoids any new HTTP surface area.
- The guard condition `len(result) == 0` is correctly placed *after* the XML unmarshalling loop, not earlier, so systems with genuine zero results (no transports at all) still trigger the fallback rather than returning a misleading empty slice.
- `transposeDataPreview` returns a valid `*QueryResult` (not nil) even when `numCols == 0`, so the fallback's column-index loop safely produces zero iterations and returns `nil, nil` — the empty-result path is handled without panics.
- Column-index discovery via a switch loop is sound: it tolerates any column ordering SAP may return, and the guard `if trkorrIdx < 0` surfaces the misconfiguration rather than silently returning empty.
- Both new unit tests (`FallsBackToE070WhenADTReturnsEmpty` and `SkipsFallbackWhenADTReturnsResults`) are end-to-end httptest tests that verify the full code path through the real `httpClient` — not just mocked internals.
- The registry test fix (`emptyTransports` now returns a real transport) is correct: the old empty-root XML would have triggered the E070 fallback inside the handler, which would have hit the catch-all `default: 200` and likely returned confusing results or panicked on the missing `datapreview` endpoint. Using a real transport avoids that.
- Investigation tests are gated with `//go:build integration` and carry thorough doc comments explaining each hypothesis — valuable for future archaeology.
- All unit tests pass (`go test -count=1 ./...`) and the integration build is clean (`go build -tags integration ./adt/...`, `go vet -tags integration ./adt/...`).

---

### Issues

#### Critical (Must Fix)

**None.**

---

#### Important (Should Fix)

**1. SQL injection guard on `user` is incomplete: lowercase letters pass through**

`strings.ContainsAny(user, "'\"\;")` blocks the obvious attack characters, and `validateSelectOnly` inside `RunQuery` blocks semicolons in the composed SQL as a second layer. However a SAP username is defined as uppercase A–Z, 0–9, and `_` (at most 12 characters, BNAME field). A value like `METZEJ OR 1=1--` passes the `ContainsAny` guard (no quotes, backslash, or semicolon) and would inject into the WHERE clause. ABAP Open SQL does not support `--` comments, and SAP will reject `OR 1=1` syntactically in the datapreview endpoint, so this is unlikely to be exploitable in practice. But it leaves the guard semantically wrong: it should *allow* only the characters a legitimate SAP username can contain, not just reject a small forbidden set. A tighter allowlist:

```go
if !regexp.MustCompile(`^[A-Z0-9_]{1,40}$`).MatchString(user) {
    return nil, fmt.Errorf("GetTransportRequests: invalid user %q", user)
}
```

(Compile once as a package-level var, not inside the function.)

The current approach is *defense in depth* — the composed SQL always starts with `SELECT`, `validateSelectOnly` strips semicolons, and ABAP Open SQL rejects arbitrary expressions — but the guard is logically inverted and could mislead future maintainers.

**2. No test for the injection-guard rejection paths**

The two new unit tests cover the happy path and the skip-fallback path. There is no test verifying that `getTransportRequestsViaQuery` returns an error when `user` contains a forbidden character (e.g. `user = "ME'TZEJ"`). Given that this is a security-relevant code path, at least one negative test would be valuable.

---

#### Minor (Nice to Have)

**3. Variable name `sql` shadows nothing but is a reserved identifier style risk**

The local variable `sql` in `getTransportRequestsViaQuery` does not shadow any import in the current file (the `database/sql` package is not imported here), so it compiles and vets clean. However the name conflicts with the conventional `database/sql` package name. Renaming to `query` or `stmt` would eliminate any future confusion if the file's import list changes.

**4. `TestGetTransportRequests_AllSystems_Integration` asserts `len > 0`**

The integration test fails with `t.Fatal("expected at least one modifiable transport, got 0")` if the system genuinely has zero open transports (e.g. a fresh or fully-released dev system). This assumption may be too strong for a CI environment. A `t.Skip` or `t.Log` with a softer assertion (validate structure when results exist, but do not fail on empty) would be more robust. The existing `TestGetTransportRequests_Integration` has the same pattern and this is consistent with how the codebase handles it elsewhere, so this is a style note rather than a blocking issue.

**5. Investigation test file name implies permanent residence**

`transport_syst_cust_investigation_integration_test.go` is a 397-line file of hypothesis-probing tests that were written to confirm the root cause (PR #66). Now that the root cause is confirmed and fixed, these tests have no ongoing regression value — they document rejected hypotheses, not the implemented fix. Keeping them adds build weight and may confuse future reviewers. Consider either deleting the file or moving it to a `testdata/` note. That said, the file is integration-only (`//go:build integration`) so it adds zero to normal CI time, making this low priority.

---

### Assessment

**GO with notes**

The core fix is correct, the fallback condition is placed right, the SQL is structurally sound for ABAP Open SQL, the LEFT OUTER JOIN degrades cleanly to an empty description when no English row exists (the `AND t.LANGU = 'E'` is part of the ON clause, not a WHERE, so non-English-only systems produce a row with `AS4TEXT = ""` rather than dropping the transport), and the `STRKORR = ''` guard correctly excludes tasks (child entries that point back to their parent request via STRKORR). The unit test coverage is solid for the happy and skip paths. The injection guard issue (#1) is the most substantive finding but is mitigated by the layered `validateSelectOnly` check in `RunQuery` and ABAP's own SQL parser. Fix the guard before merge if you can; it does not block the PR functionally.

Jonatan-HF:
## Review fixes applied (9674539)

Both Important issues from the review addressed:

1. **Allowlist instead of blocklist** — replaced `strings.ContainsAny(user, "'\"\;")` with `validTransportUserParam` (exported helper): allows only `[A-Za-z0-9_.-]` up to 40 chars. The error message now names the constraint explicitly.

2. **Error-path test** — added `TestGetTransportRequests_RejectsInvalidUserInFallback`: mocks ADT to return empty root (triggering fallback), passes `user = "ME'TZEJ"`, asserts non-nil error.

Minor items:
- Renamed `sql` → `query` locally.
- Investigation test file kept (integration-only, no CI cost, documents rejected hypotheses for posterity).
- Integration test fragility left as-is — consistent with existing pattern; a system with zero open transports isn't a valid test target.

Jonatan-HF:
## Integration test results

### S4U (S/4HANA — KORRDEV=SYST/CUST, the bug system)

**PASS** — 916 modifiable transport requests returned via E070 fallback.
Prior to this fix: 0 results.

```
TestGetTransportRequests_Integration          PASS  (1.28s)
TestGetTransportRequests_AllSystems/S4U       PASS  (1.25s)
```

Sample results (owners/descriptions redacted):
- 916 transports returned, mix of SYST/CUST/unnamed KORRDEV types
- Number, owner, status all populated correctly
- Description empty (expected: data preview rejects JOINs on S4U — see commit 68681ee)

### HFQ (R/3/ECC — KORRDEV=K, normal path)

**PASS** — 4 modifiable transport requests returned directly via ADT endpoint (fallback not triggered).

```
TestGetTransportRequests_AllSystems/HFQ       PASS  (0.70s)
```

### Fix iteration during integration run

The initial implementation used `LEFT OUTER JOIN E07T` to fetch descriptions.
S4U's data preview endpoint rejected it with `ExceptionDataPreviewGeneral: Only one SELECT statement is allowed` — the endpoint does not support JOINs on this S4U release.
Fix (68681ee): dropped to single-table `SELECT TRKORR, AS4USER, TRSTATUS FROM E070`.
Description is now empty in the fallback path (callers can use `GetTransportInfo` per transport if needed).
All unit tests updated accordingly.

### Cleanup

No objects left in `$TMP`. The setup fixture transport `S4UK903273` was created by `TestMain` (standard fixture pattern) and will persist until the next full test run cleans it up.

---

**Verdict: ready to merge.** Removing `needs:integration-test` label.

adt/transport.go
```diff
@@ -393,6 +393,96 @@ func (c *httpClient) GetTransportRequests(ctx context.Context, user, status stri
...
+	if len(s) > 40 {
+		return false
+	}
+	for _, r := range s {
+		if ('A' <= r && r <= 'Z') || ('a' <= r && r <= 'z') || ('0' <= r && r <= '9') || r == '_' || r == '-' || r == '.' {
+			continue
+		}
+		return false
+	}
+	return true
```
hf-kklein: why don't we use a regex? compile the pattern once
https://pkg.go.dev/regexp#MustCompile

then use `.MatchString` function.
https://pkg.go.dev/regexp#example-package

if there's a good reason to not use regex, please explain in a comment but imho it's the right tool for this purpose.

adt/transport.go
```diff
@@ -393,6 +393,96 @@ func (c *httpClient) GetTransportRequests(ctx context.Context, user, status stri
...
+}
+
+func (c *httpClient) getTransportRequestsViaQuery(ctx context.Context, user, status string) ([]TransportRequest, error) {
+	if user != "" && !validTransportUserParam(user) {
+		return nil, fmt.Errorf("GetTransportRequests: user %q contains characters not allowed in a transport query", user)
+	}
+	if status != "" && (len(status) != 1 || status[0] < 'A' || status[0] > 'Z') {
+		return nil, fmt.Errorf("GetTransportRequests: status must be a single uppercase letter, got %q", status)
+	}
```
hf-kklein: what's the second part of this condition? status not empty AND ...? maybe move it to a readable variable:

```go
isFooBar := len(status) != 1 || status[0] < 'A' || status[0] > 'Z'
if status!="" && isFooBar{
...
```

adt/transport.go
```diff
@@ -393,6 +393,96 @@ func (c *httpClient) GetTransportRequests(ctx context.Context, user, status stri
...
+		return false
+	}
+	return true
+}
+
+func (c *httpClient) getTransportRequestsViaQuery(ctx context.Context, user, status string) ([]TransportRequest, error) {
+	if user != "" && !validTransportUserParam(user) {
+		return nil, fmt.Errorf("GetTransportRequests: user %q contains characters not allowed in a transport query", user)
+	}
+	if status != "" && (len(status) != 1 || status[0] < 'A' || status[0] > 'Z') {
```
hf-kklein: can't we use a regex here too? like in validTransportUserParam? that'd make it much more obvious imo.

hf-kklein (approved):
lgtm.

adt/transport.go
Jonatan-HF: Done in 64779d3 — replaced the character loop with a package-level `transportUserRe = regexp.MustCompile(`^\[A-Za-z0-9_.-\]{1,40}$`)`. Compiled once, used via `.MatchString`.

adt/transport.go
Jonatan-HF: Done in 64779d3 — same pattern: `transportStatusRe = regexp.MustCompile(`^\[A-Z\]$`)`, package-level, used via `.MatchString`. The byte-comparison is gone.

adt/transport.go
Jonatan-HF: Done in 64779d3 — replaced the entire condition with `!transportStatusRe.MatchString(status)` where `transportStatusRe = regexp.MustCompile(`^\[A-Z\]$`)`. No intermediate variable needed — the regex intent is self-documenting.

PR #76, opened 2026-06-18
feat: expose ObjectURI(objectType, name) helper

### LINKED ISSUE #75: feat: expose ObjectURI(objectType, name) helper
## Summary

Expose a public helper that builds the ADT resource URI for an object from its type + name, backed by the existing (private) `objectTypeMap`.

```go
func ObjectURI(objectType, name string) (string, error)
```

e.g. `ObjectURI("PROG", "ZFOO") == "/sap/bc/adt/programs/programs/zfoo"`. Name is lower-cased to match the ADT REST URI convention; unknown types return an error listing supported types.

## Motivation

The consumer (mcp-server-abap) currently reimplements subsets of this object-type → endpoint mapping in two places (`tools/rollback.go`, `tools/verify.go`), duplicating knowledge that `objectTypeMap` already owns. Exposing `ObjectURI` lets those duplicates be deleted.

Consumer issue: Hochfrequenz/aibap.mcp#410

hf-kklein:
## What

Adds a public `ObjectURI(objectType, name string) (string, error)` helper that builds the ADT resource URI for an object from its type + name, backed by the existing private `objectTypeMap`.

```go
adt.ObjectURI("PROG", "ZFOO") // -> "/sap/bc/adt/programs/programs/zfoo", nil
```

- Name is lower-cased to match the ADT REST URI convention.
- Object type is matched case-insensitively.
- Unknown types return an error listing the supported types.

Also extracts the supported-types listing into `supportedObjectTypes()` (sorted, deterministic) so the error message is shared between `ObjectURI` and `CreateObject` rather than duplicating the map-iteration block — pre-empts a `dupl` finding.

## Why

mcp-server-abap reimplements subsets of this object-type → endpoint mapping in two places (`tools/rollback.go`'s `sourceTypeToEndpoint`, and a hardcoded literal in `tools/verify.go`). Exposing `ObjectURI` lets those duplicates be deleted on the next bump.

Closes #75. Consumer: Hochfrequenz/aibap.mcp#410

## Tests

- `adt/object_uri_test.go` — table-driven coverage of all 9 mapped types, case-insensitive type, already-lower name, and the unsupported-type error (empty return + type named + supported list present).
- `go test ./...` green; `go vet ./...` clean; `go build -tags integration ./adt/...` compiles.

No integration test: `ObjectURI` is a pure string function with no HTTP/SAP interaction, so the `eachSystem(t)` pattern does not apply. The URIs it produces are already exercised against live SAP via `CreateObject`/`GetSource` etc., which read from the same `objectTypeMap`.

## Lint note

`golangci-lint` is not installed in my environment so I could not run it locally; the `supportedObjectTypes()` extraction was done specifically to avoid the `dupl` duplication the two identical list-builders would otherwise trigger.

### DIFF
--- a/adt/object.go
+++ b/adt/object.go
@@ -49,6 +49,10 @@ func supportedObjectTypes() []string {
 // name, e.g. ObjectURI("PROG", "ZFOO") == "/sap/bc/adt/programs/programs/zfoo".
 // The name is lower-cased to match the ADT REST URI convention. It returns an
 // error for object types not known to objectTypeMap.
+//
+// The name is appended verbatim (after lower-casing) and is not URL-encoded, so
+// callers passing namespaced names (e.g. "/NS/NAME") are responsible for any
+// required escaping; an empty name yields a trailing slash.
 func ObjectURI(objectType, name string) (string, error) {
 	info, ok := objectTypeMap[strings.ToUpper(objectType)]
 	if !ok {
--- a/adt/object.go
+++ b/adt/object.go
@@ -5,6 +5,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"net/http"
+	"sort"
 	"strings"
 
 	"github.com/Hochfrequenz/adtler/adt/adtxml"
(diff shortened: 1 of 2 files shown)


hf-kklein:
**GO with notes** — independent review.

## Verdict
Correct, well-tested, idiomatic, and it does what consumer aibap.mcp#410 needs. Safe to merge. Two non-blocking observations below.

## Build & test (worktree on the PR commit `4fa7a79`)
- `go test ./...` → all packages **ok** (adt, adtxml, custexport, auth).
- `go vet ./...` → clean.
- `go build -tags integration ./adt/...` → compiles, exit 0.

## Correctness
- URI shape is right. `objectTypeMap` endpoints have **no** trailing slash; `ObjectURI` inserts the `/` separator and lower-cases the name, so `ObjectURI("PROG","ZFOO")` == `/sap/bc/adt/programs/programs/zfoo` — byte-identical to today’s rollback.go (`"/sap/bc/adt/programs/programs/" + strings.ToLower(name)`). The trailing-slash mismatch flagged in #410 is handled correctly.
- Case handling is correct ADT convention: type matched case-insensitively (`ToUpper`), name lower-cased. Verified against the codebase — `CreateObject` POSTs to the bare collection endpoint and SAP assigns the resource; every other working call site (e.g. the PR#407 integration test deletes an uppercase-created PROG via `strings.ToLower(objName)`, and the `z_adt_mcp_test_*` integration URIs) addresses the resource by **lowercase** name. So lowercase is the canonical server form.

## The verify.go case-change question (the important one)
`tools/verify.go` currently builds `"/sap/bc/adt/programs/programs/" + name` where `name` is **UPPERCASE** `Z_MCP_VERIFY_NNNNNN`. Switching to `adt.ObjectURI("PROG", name)` lower-cases it. This **is** a produced-URI change (uppercase → lowercase), but it is **safe**: ADT program resource URIs are case-insensitive on the path and the canonical form is lowercase — confirmed by the fact that aibap.mcp already creates objects with uppercase names and then locks/sets-source/deletes them via lowercase URIs throughout (PR#407 test line ~230, integration suite). The temp object is created via `CreateObject(ctx,"PROG",name,...)` which doesn’t embed the name in the request path, so the only consumer of `objectURI` is the subsequent lock/getsource/setsource/syntaxcheck/delete chain — all of which work with the canonical lowercase form. Net: the change moves verify.go onto the same canonical form the rest of the codebase already relies on. No behavior risk.

## supportedObjectTypes() refactor
- Preserves `CreateObject` behavior exactly (same map lookup, same error string template). Only difference: the supported-type list is now **sorted/deterministic**. No unit or integration test asserts on the order of that error message (checked both repos), so no test breaks; this is strictly better (avoids flaky diffs and pre-empts `dupl`). Good call.

## Test coverage
Solid: all 9 mapped types, case-insensitive type, already-lower name, and the unsupported-type error (empty return + type named + supported list present). Two gaps, both **pre-existing and out of scope**, worth a one-line doc note rather than code:
1. **Empty name** → `/sap/bc/adt/programs/programs/` (dangling slash). Not validated. Old code didn’t validate either; in-scope callers always pass a real name.
2. **Namespaced names** (`/NS/NAME`) → lower-cased to `/ns/name`, yielding `.../programs//ns/name`. Real ADT URL-encodes namespace slashes (`%2f`). Neither rollback.go nor verify.go handled this before, so no regression — but if `ObjectURI` is meant to be the canonical builder, a follow-up to either handle or explicitly reject `/`-containing names (and a doc sentence saying namespaced names are unsupported) would prevent a future caller getting a silently-wrong URI.

## Conclusion
Merges cleanly into the #410 refactor (rollback.go can drop `sourceTypeToEndpoint` and call `adt.ObjectURI`; verify.go replaces its literal). Approve. Consider a doc-comment note on the namespaced-name limitation.

PR #80, opened 2026-06-18
feat: export TextElementLockURI(objectURI) helper

### LINKED ISSUE #79: feat: export TextElementLockURI(objectURI) helper
## Summary

Export a public helper that returns the textelements resource URI an object's text symbols/selection texts are bound to:

```go
func TextElementLockURI(objectURI string) (string, error)
```

Thin wrapper over the existing (private) `resolveTextElementPath`. Also neutralizes that function's misleading `"GetTextElements: ..."` error prefix (it is shared by GetTextElements, SetTextElements, and now this helper).

## Motivation

`SetTextElements` resolves this path internally for the write but does not expose it, so the consumer (mcp-server-abap) reimplements the program/class/function-group → `/sap/bc/adt/textelements/...` rewrite in `tools/textelements.go` (`textElementsResourceURI`) purely to know which enqueue resource to lock. SAP binds the text-element lock to that resource, not the object. Exposing `TextElementLockURI` lets the duplicate be deleted. The two copies have already diverged (adtler matches prefixes case-insensitively; the consumer copy is case-sensitive).

Consumer issue: Hochfrequenz/aibap.mcp#411

hf-kklein:
## What

- Adds `TextElementLockURI(objectURI string) (string, error)` — a thin public wrapper over the private `resolveTextElementPath`, returning the `/sap/bc/adt/textelements/...` resource URI an object's text elements are bound to.
- Neutralizes `resolveTextElementPath`'s `"GetTextElements: ..."` error prefix (shared by GetTextElements, SetTextElements, and the new helper).

## Why

`SetTextElements` resolves this path internally but doesn't expose it, so mcp-server-abap reimplements the program/class/function-group → textelements rewrite (`textElementsResourceURI`) just to know which enqueue resource to lock — SAP binds the text-element lock to that resource, not the object. The two copies have already diverged (adtler is case-insensitive; the consumer copy is case-sensitive). Exposing this lets the duplicate be deleted.

Closes #79. Consumer: Hochfrequenz/aibap.mcp#411

## Tests

- `adt/textelements_test.go::TestTextElementLockURI` — asserts the wrapper returns the same textelements URI SetTextElements writes to (program case) and errors on an unsupported object type. The existing `TestResolveTextElementPath` (covering classes/function groups) still passes.
- `go test ./...` green; `go vet` clean; `golangci-lint --enable dupl,goconst,gocyclo` → 0 issues; `go build -tags integration ./adt/...` compiles.

No integration test: the helper is a pure URI derivation with no HTTP; the resulting URIs are exercised against live SAP via the existing SetTextElements/GetTextElements integration tests.

## API note

Additive export only — no interface change (it's a package-level function, not a method), so no implementers/mocks are affected.

### DIFF
--- a/adt/textelements.go
+++ b/adt/textelements.go
@@ -136,6 +136,15 @@ func (c *httpClient) writeTextElementSource(ctx context.Context, path, contentTy
 	return checkResponse(resp)
 }
 
+// TextElementLockURI returns the ADT textelements resource URI that the text
+// symbols and selection texts of the given object are bound to. SAP attaches
+// the enqueue lock for text elements to this resource rather than to the object
+// itself, so callers that lock before SetTextElements must acquire the lock on
+// this URI. The objectURI must be a program, class, or function group URI.
+func TextElementLockURI(objectURI string) (string, error) {
+	return resolveTextElementPath(objectURI)
+}
+
 func resolveTextElementPath(objectURI string) (string, error) {
 	upper := strings.ToUpper(objectURI)
 	for prefix, tePath := range textElementEndpoints {
@@ -145,7 +154,7 @@ func resolveTextElementPath(objectURI string) (string, error) {
 			return tePath + name, nil
 		}
 	}
-	return "", fmt.Errorf("GetTextElements: unsupported object type for URI %q (only programs, classes, function groups)", objectURI)
+	return "", fmt.Errorf("unsupported object type for text elements: URI %q (only programs, classes, function groups)", objectURI)
 }
 
 func (c *httpClient) readTextElementSource(ctx context.Context, path, accept string) (string, error) {
--- a/adt/textelements_test.go
+++ b/adt/textelements_test.go
@@ -73,6 +73,22 @@ func TestResolveTextElementPath(t *testing.T) {
 	}
 }
 
+func TestTextElementLockURI(t *testing.T) {
+	// The public wrapper must return the same textelements resource URI that
+	// SetTextElements writes to, so callers lock the correct enqueue resource.
+	got, err := TextElementLockURI("/sap/bc/adt/programs/programs/ZTEST")
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if want := "/sap/bc/adt/textelements/programs/ZTEST"; got != want {
+		t.Errorf("TextElementLockURI = %q, want %q", got, want)
+	}
+
+	if _, err := TextElementLockURI("/sap/bc/adt/ddic/tables/ZTABLE"); err == nil {
+		t.Error("expected error for unsupported object type")
+	}
+}
+
 func TestFormatTextSymbols(t *testing.T) {
 	symbols := []TextSymbol{
 		{Key: "001", Text: "Hello World", MaxLength: 50},
--- a/adt/textelements_test.go
+++ b/adt/textelements_test.go
@@ -84,6 +84,17 @@ func TestTextElementLockURI(t *testing.T) {
 		t.Errorf("TextElementLockURI = %q, want %q", got, want)
 	}
 
+	// The prefix is matched case-insensitively (the original-case name suffix
+	// is preserved). This is the behavior that differs from a naive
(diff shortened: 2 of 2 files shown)


hf-kklein:
## Review verdict: GO

Reviewed on the merits with no prior context. Branch \`feat/textelement-lock-uri\` was already checked out in the main adtler dir; I reviewed there directly (no worktree created/needed).

### Local checks (all green)
- \`go test ./...\` → ok (adt, adtxml, custexport, auth)
- \`go vet ./...\` → clean
- \`gofmt -l adt/\` → no output
- \`golangci-lint run --enable dupl,goconst,gocyclo ./...\` → 0 issues
- \`go build -tags integration ./adt/...\` → compiles

### Correctness — lock URI == write URI
Confirmed. \`TextElementLockURI\` is a one-line wrapper that calls \`resolveTextElementPath\` (textelements.go:144-146), the *same* resolver \`SetTextElements\` (line 79) and \`GetTextElements\` (line 41) use. The lock URI and the write URI are derived by identical code — they cannot drift. This is exactly the contract #79/#411 need: the consumer locks this URI, then SetTextElements writes to it.

### Case-sensitivity behavior for the #411 bump
The consumer's current \`textElementsResourceURI\` is case-SENSITIVE (\`strings.HasPrefix\`); adtler is case-INSENSITIVE (\`strings.ToUpper\` prefix match). Walk-through:

- Both implementations preserve the **original-case name suffix** and emit a **fixed lowercase** textelements prefix. adtler returns \`tePath + objectURI[len(prefix):]\`; the consumer does \`strings.Replace(lowercasePrefix → lowercaseReplacement, 1)\`.
- For a typical lowercase URI \`/sap/bc/adt/programs/programs/zfoo\` → both return \`/sap/bc/adt/textelements/programs/zfoo\`. **Byte-identical.** Same for canonical mixed-case names like \`.../programs/ZTEST\` → \`.../textelements/programs/ZTEST\`.
- They diverge ONLY when the *prefix portion* is not exactly lowercase (e.g. \`/SAP/BC/ADT/programs/programs/ZFOO\`): adtler matches and returns a canonical-lowercase-prefix URI; the consumer returns the "unsupported" error.
- Net: adtler is strictly **more permissive** on the prefix, never less. There is no input for which adtler returns a *different non-error* URI than the consumer. Real MCP traffic uses canonical lowercase ADT paths (from adtler's own search/info responses), so the divergence is not hit in practice. The bump PR should note this as a behavior *widening*, not a regression.

### Error-message change
\`resolveTextElementPath\` error prefix changed \`"GetTextElements: ..."\` → \`"unsupported object type for text elements: ..."\`. Grepped **both** repos for \`GetTextElements:\` and the old-string variants — no test or caller asserts on the old text in adtler or aibap.mcp. Safe. The consumer's own \`textElementsResourceURI\` produces yet a third message ("text elements not supported for ..."), but #411 deletes that path entirely, so no assertion conflict.

### Test quality
\`TestTextElementLockURI\` covers the program happy path (asserts exact URI == SetTextElements target) and the unsupported-type error. \`TestResolveTextElementPath\` still covers class + function-group paths. Adequate. Minor (non-blocking) note: the new test only exercises the \`programs\` branch directly via the public wrapper, and neither test asserts the case-insensitive match (e.g. \`/sap/bc/adt/PROGRAMS/programs/ZTEST\`) — which is the one behavior that differs from the consumer copy and the whole reason the duplicate "already diverged." A one-line mixed-case case in the table would lock in that contract. Not required for GO.

### API impact
Confirmed additive: package-level function \`TextElementLockURI\`, not a method on any interface. No implementers/mocks affected — unlike prior interface-changing PRs.

### #79 / #411
Satisfies #79 (exports the wrapper, neutralizes the misleading error prefix). Unblocks #411: the consumer can delete \`textElementsResourceURI\` and call \`adt.TextElementLockURI\`. Acceptance criteria are met on the adtler side. Per #411's own note, re-run the \`set_text_elements\` integration test on bump to confirm the lock still lands on the textelements resource.

**GO** — merge-ready pending CI + the standard real-SAP reproducer/integration check on the consumer bump.

PR #82, opened 2026-06-18
feat: ErrorKind + ClassifyError; export remaining ExceptionType constants

hf-kklein:
## What

1. **Export 8 more `ExceptionType*` constants** in `adt/types.go` (AlreadyExists, LockConflict, InvalidEtag, NotAcceptable, UnsupportedMediaType, UnprocessableEntity, NotAllowed, CreationFailure), each annotated with its cross-system HTTP status.
2. **Add `ErrorKind` + `ClassifyError`** (`adt/errorkind.go`): `ClassifyError(err) ErrorKind` maps an `*ADTError` to a stable kind — by SAP `Type` first (language/status-independent), then by HTTP status as a fallback. `ErrorKind.String()` gives stable lowercase ids for logging/keying.

## Why

mcp-server-abap's `tools/errors.go` redeclares these 8 constants and encodes the cross-system `(Type, status) → meaning` tiering locally. That is SAP protocol knowledge and belongs in adtler; the MCP layer should keep only the hint *wording* (which names MCP tools), keyed by `ErrorKind`.

Closes #81. Consumer: Hochfrequenz/aibap.mcp#413 (the consumer migration is coordinated with aibap.mcp#363 — this PR is the adtler-side API only).

## Design decisions (please scrutinise)

- **Type precedence:** `ClassifyError` consults `Type` before status, so R/3's typed 405-AlreadyExists → `ErrorAlreadyExists`, not `ErrorMethodNotAllowed`. Unrecognised/empty Type falls through to the status switch.
- **Bare 423 → `ErrorLocked`**, deliberately distinct from the internal `isInvalidLockHandle` heuristic (typeless 423 = invalid lock handle, for SetSource's delivery-mechanism retry). `isInvalidLockHandle` is intentionally left untouched — wiring it to ClassifyError would regress that retry.
- **`ExceptionTypeResourceWrongData`** (pre-existing constant) → `ErrorUnprocessable`.
- Split into `classifyByExceptionType` + `classifyByStatusCode` to keep each function's cyclomatic complexity well under the gocyclo threshold.

## Tests

`adt/errorkind_test.go`: every Type→kind, every status→kind (incl. an unmapped status → Unknown), Type-precedence-over-status, unrecognised-Type-falls-back-to-status, bare-423→Locked, non-ADTError/nil→Unknown, wrapped (`%w`) ADTError, and `String()` values.

[...]

PR #84, opened 2026-06-18
feat: ReleaseTransportVerified — detect ECC silent-fail release

### LINKED ISSUE #83: feat: ReleaseTransportVerified — detect ECC silent-fail release
## Summary

Add `ReleaseTransportVerified(ctx, transportNumber string, includeTasks bool) (*ReleaseResult, error)`, which releases a transport (optionally including tasks) and then **confirms** the outcome by re-reading the request status.

Some systems — notably ECC — return HTTP 200 from the release endpoint while leaving the request modifiable. The plain `ReleaseTransport` only inspects the release *report* (`checkReleaseResponse`) and reports success in that case. The verified variant re-reads the status: a request still in the modifiable (`"D"`) state is reported as `Released=false` (with a nil error) so callers can fall back to another release mechanism.

Also export `TransportStatusModifiable = "D"` / `TransportStatusReleased = "L"` constants.

## Motivation

mcp-server-abap's `release_transport` handler performs exactly this post-release `GetTransportInfo` + `Status == "D"` check inline. That's ADT response semantics and belongs in adtler, beside `checkReleaseResponse`. The BlackMagic fallback *routing* stays in the MCP layer.

## API shape

Additive (non-breaking): a new method + `ReleaseResult` type, rather than changing `ReleaseTransport`'s `error`-only signature. Added to the `TransportClient` interface, `httpClient`, and `ClientRegistry`.

Consumer issue: Hochfrequenz/aibap.mcp#414

hf-kklein:
## What

- Adds `ReleaseTransportVerified(ctx, transportNumber string, includeTasks bool) (*ReleaseResult, error)`: releases (optionally with tasks), then re-reads the request status to confirm. A request still modifiable (`"D"`) after a 2xx release → `Released=false` with a **nil error** (the ECC "silent failure"); a genuine release error is returned unchanged; an unreadable post-release status is treated optimistically as `Released=true` (matching prior caller behavior).
- Adds `ReleaseResult{Transport, Released}`.
- Exports `TransportStatusModifiable = "D"` / `TransportStatusReleased = "L"`.

## Why

mcp-server-abap's `release_transport` handler does this exact post-release `GetTransportInfo` + `Status == "D"` check inline — ADT response semantics that belong here beside `checkReleaseResponse`. The BlackMagic fallback *routing* stays MCP-side.

Closes #83. Consumer: Hochfrequenz/aibap.mcp#414.

## API shape (note for reviewers)

Chose the **additive** path: a new method + result type rather than changing `ReleaseTransport`'s `error`-only signature (which would break all callers). `ReleaseTransportVerified` is added to the exported `TransportClient` interface — the only in-repo full-`adt.Client` implementer outside the real clients is the custexport test mock, updated here. The consumer adapts its mock in the #414 bump.

`includeTasks bool` is a single param rather than a second `…WithTasksVerified` method — it composes over the existing `ReleaseTransport`/`ReleaseTransportWithTasks` split internally.

## Tests

`adt/release_verified_test.go`: synchronous release then status read returning `L` (Released=true), `D` (silent-fail Released=false), and an unreadable status (optimistic Released=true); plus release-error-propagates (nil result + error).

No integration test added here: the verified flow composes `ReleaseTransport` + `GetTransportInfo`, both already integration-tested; the silent-fail path is ECC-specific and is exercised by the #414 consumer reproducer on bump.

### DIFF
--- a/adt/client.go
+++ b/adt/client.go
@@ -115,6 +115,7 @@ type TransportClient interface {
 	DeleteTransport(ctx context.Context, transportNumber string) error
 	ReleaseTransport(ctx context.Context, transportNumber string) error
 	ReleaseTransportWithTasks(ctx context.Context, transportNumber string) error
+	ReleaseTransportVerified(ctx context.Context, transportNumber string, includeTasks bool) (*ReleaseResult, error)
 	GetTransportRequests(ctx context.Context, user, status string) ([]TransportRequest, error)
 	AddToTransport(ctx context.Context, objectURI, transport string) error
 	RemoveFromTransport(ctx context.Context, taskNumber, parentTransport, pgmID, objectType, objectName, wbType, position string) error
--- a/adt/custexport/export_test.go
+++ b/adt/custexport/export_test.go
@@ -133,6 +133,9 @@ func (m *mockClient) DeleteTransport(context.Context, string) error {
 func (m *mockClient) ReleaseTransport(context.Context, string) error {
 	panic("not implemented")
 }
+func (m *mockClient) ReleaseTransportVerified(context.Context, string, bool) (*adt.ReleaseResult, error) {
+	panic("not implemented")
+}
 func (m *mockClient) ReleaseTransportWithTasks(context.Context, string) error {
 	panic("not implemented")
 }
--- a/adt/registry.go
+++ b/adt/registry.go
@@ -179,6 +179,9 @@ func (r *ClientRegistry) RemoveFromTransport(ctx context.Context, taskNumber, pa
 func (r *ClientRegistry) ReleaseTransportWithTasks(ctx context.Context, transportNumber string) error {
 	return r.activeClient().ReleaseTransportWithTasks(ctx, transportNumber)
 }
+func (r *ClientRegistry) ReleaseTransportVerified(ctx context.Context, transportNumber string, includeTasks bool) (*ReleaseResult, error) {
+	return r.activeClient().ReleaseTransportVerified(ctx, transportNumber, includeTasks)
+}
 func (r *ClientRegistry) GetTransportInfo(ctx context.Context, transportNumber string) (*TransportRequest, error) {
 	return r.activeClient().GetTransportInfo(ctx, transportNumber)
 }
--- a/adt/release_verified_test.go
+++ b/adt/release_verified_test.go
@@ -1,0 +1,102 @@
+package adt_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/Hochfrequenz/adtler/adt"
+	sapmcpconfig "github.com/Hochfrequenz/sap-mcp-config"
+)
+
+// releaseVerifiedServer mocks a synchronous release followed by a status read.
+// postReleaseStatus is the tm:status attribute returned by the status GET; pass
+// "" to make the status read fail (no <request> element).
+func releaseVerifiedServer(t *testing.T, transport, postReleaseStatus string) *httptest.Server {
+	t.Helper()
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch {
+		case r.URL.Path == csrfEndpoint:
+			w.Header().Set("X-CSRF-Token", "token")
+			w.WriteHeader(http.StatusOK)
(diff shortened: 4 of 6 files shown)


hf-kklein:
## Review verdict: GO with notes

Reviewed independently against #83, consumer #414, and the consumer's inline `release_transport` logic (`aibap.mcp/tools/transport.go:165-197`). Local checks all green on this branch (Go 1.25.0):

- `go test ./...` — ok (all 4 packages)
- `go test -run ReleaseTransportVerified ./adt/...` — ok
- `go vet ./...` — clean
- `gofmt -l adt/` — clean (no files)
- `go build -tags integration ./adt/...` — compiles

### Semantic fidelity vs the consumer's inline logic — EXACT
Walked every branch of `ReleaseTransportVerified` (`adt/transport.go:200-221`) against the consumer's `silentFail` logic:

| Condition | Consumer | adtler | Match |
|---|---|---|---|
| `releaseErr != nil` | not silentFail; surfaces error | returns `(nil, err)` unchanged | ✓ |
| release ok, `Status == "D"` | `silentFail=true` → fallback | `Released=false`, nil err | ✓ |
| release ok, `GetTransportInfo` errors | silentFail stays false → released | `Released=true` (optimistic) | ✓ |
| release ok, `Status == "L"` | released | `Released=true` | ✓ |

The guard `infoErr == nil && info != nil && info.Status == TransportStatusModifiable` is a character-for-character port of the consumer's condition. The BlackMagic fallback *routing* correctly stays consumer-side; adtler returns only the verified outcome. This satisfies #83 and unblocks #414.

### Optimistic fallback on status-read failure — acceptable, with a caveat
Treating an unreadable post-release status as `Released=true` matches prior caller behavior, and is the conservative choice for the *consumer* (a read failure won't spuriously route a successful S/4 release to the fallback). The downside: a transient network/parse failure on the status GET masks a genuine ECC silent-fail, reporting success when the transport is still modifiable. This is a pre-existing behavior being faithfully preserved, not a regression, and the doc comment is honest about it. Fine to keep; worth a follow-up if false-negatives are observed live.

### includeTasks plumbing — correct
`true` → `ReleaseTransportWithTasks`, `false` → `ReleaseTransport`; composes over the existing split rather than a parallel `…WithTasksVerified`. Good additive shape.

### Constants — correct
`TransportStatusModifiable="D"` / `TransportStatusReleased="L"` (`adt/types.go`) match the long-standing `TransportRequest.Status` comment and the consumer's literals. Used in the verified guard. `TransportStatusReleased` is referenced only in tests so far — fine.

### Interface/mock impact — verified across both repos
- adtler implementers of the new `TransportClient.ReleaseTransportVerified` all compile: `httpClient` (`adt/transport.go`), `ClientRegistry` delegation (`adt/registry.go:182`), custexport test mock (`adt/custexport/export_test.go`).
- **Consumer mock for the #414 bump:** exactly one `adt.Client` implementer needs the method added — `mockClient` in `aibap.mcp/tools/source_test.go:30` (alongside its existing `ReleaseTransport` at :255 and `ReleaseTransportWithTasks` at :261). The `mockBlackMagic*` types implement the fallback interface, not `adt.Client`, so they are unaffected.

### Test quality — covers silent-fail detection, not just happy path
`release_verified_test.go` table covers L (Released=true), D (silent-fail Released=false), unreadable status (optimistic true), plus error-propagation (nil result + error). The httptest mock is faithful: synchronous `newreleasejobs` release report → `checkReleaseResponse` passes → status GET returns the configurable `tm:status`. The unreadable case omits `<tm:request>`, correctly driving `parseTransportInfo` into its 'no request element' error → the `infoErr != nil` path. Not covered: the async-polling release branch and `includeTasks=true` dispatch — acceptable, as both delegate to already-tested methods and the live ECC path is deferred to the #414 reproducer per the PR note.

### Note (non-blocking): branch is behind main / carries duplicate commits
The PR branch's merge-base is `2aaee63`, but #82 (ErrorKind + ClassifyError, `adt/errorkind.go`, ExceptionType constants) has since merged to `main`. This branch re-introduces those same commits (`7b1ea8e`, `d4033ce`), so `gh pr diff 84` shows ~290 lines of ErrorKind noise that are unrelated to #83. Content is identical to what's on main, so this should reconcile on merge, but please **rebase on main** before merging so the PR diff reflects only the ReleaseTransportVerified change (and to avoid a confusing history). This is the only reason this is 'GO with notes' rather than a clean GO — the #83 change itself is correct and complete.

PR #86, opened 2026-06-18
feat: RollbackTransport — restore a transport's objects to their pre-transport version

### LINKED ISSUE #85: feat: RollbackTransport — restore a transport's objects to their pre-transport version
## Summary

Add `RollbackTransport(ctx, transportNumber string) (*RollbackResult, error)` plus `RollbackResult`/`RollbackEntry` types. For each R3TR PROG/CLAS/INTF/FUGR object in the transport it reads the version history, finds the version immediately before the transport, and restores that source (lock → set → activate). Non-R3TR entries and non-source object types (TABL, DTEL, …) are skipped; objects created by the transport (no prior version) are reported as failed.

The version-walking heuristic is extracted as `findPreTransportVersion` and unit-tested in isolation.

## Motivation

mcp-server-abap's `rollback_transport` tool implements this entire multi-call workflow (`doRollback`/`rollbackObject`) in the MCP layer — the R3TR/source-type filter, the "version after the transport entry = pre-transport state" heuristic, and the lock/etag/set/activate dance. That is SAP behaviour built from adtler primitives and belongs here. The MCP layer keeps only the destructive-action elicitation and result shaping.

Consumer issue: Hochfrequenz/aibap.mcp#416

hf-kklein:
## What

Adds `RollbackTransport(ctx, transportNumber string) (*RollbackResult, error)` and the `RollbackResult`/`RollbackEntry` types (`adt/rollback.go`). For each **R3TR** PROG/CLAS/INTF/FUGR object in the transport it reads the version history, finds the version immediately before the transport, and restores that source (lock → get-etag → set → activate). Non-R3TR entries and non-source object types (TABL, DTEL, …) are **skipped**; objects created by the transport (no prior version) are reported as **failed**.

The version-walking heuristic is extracted as `findPreTransportVersion` (history is newest-first; the entry after the transport's own version is the pre-transport state).

Added to the `TransportClient` interface, `ClientRegistry`, and the custexport test mock.

## Why

mcp-server-abap's `rollback_transport` tool implements this whole multi-call workflow inline (`doRollback`/`rollbackObject`). The SAP-specific parts — the R3TR/source-type filter, the version heuristic, the edit/activate dance — are built from adtler primitives and belong here. The MCP layer keeps only the destructive-action elicitation and result shaping.

Closes #85. Consumer: Hochfrequenz/aibap.mcp#416.

## Tests

- `TestFindPreTransportVersion` (pure, the SAP-specific heuristic): version-after-transport, transport spanning multiple versions, transport-not-in-history, object-created-by-transport.
- `TestRollbackTransport_SkipsNonRestorable` (httptest): a transport with a LIMU entry (→ "not R3TR") and an R3TR TABL (→ "non-source object type") — both skipped, no restore work.

No full-restore httptest: the lock→set→activate dance composes already-integration-tested primitives (`GetVersionSource`, `LockObject`, `SetSource`, `ActivateObjects`); the end-to-end restore is exercised by the #416 consumer reproducer against a live system on bump.

## Interface note

`RollbackTransport` is added to the exported `TransportClient` interface — the consumer's `mockClient` (the only full-`adt.Client` double on that side) will need the method in the #416 bump PR.

### DIFF
--- a/adt/client.go
+++ b/adt/client.go
@@ -116,6 +116,7 @@ type TransportClient interface {
 	ReleaseTransport(ctx context.Context, transportNumber string) error
 	ReleaseTransportWithTasks(ctx context.Context, transportNumber string) error
 	ReleaseTransportVerified(ctx context.Context, transportNumber string, includeTasks bool) (*ReleaseResult, error)
+	RollbackTransport(ctx context.Context, transportNumber string) (*RollbackResult, error)
 	GetTransportRequests(ctx context.Context, user, status string) ([]TransportRequest, error)
 	AddToTransport(ctx context.Context, objectURI, transport string) error
 	RemoveFromTransport(ctx context.Context, taskNumber, parentTransport, pgmID, objectType, objectName, wbType, position string) error
--- a/adt/custexport/export_test.go
+++ b/adt/custexport/export_test.go
@@ -136,6 +136,9 @@ func (m *mockClient) ReleaseTransport(context.Context, string) error {
 func (m *mockClient) ReleaseTransportVerified(context.Context, string, bool) (*adt.ReleaseResult, error) {
 	panic("not implemented")
 }
+func (m *mockClient) RollbackTransport(context.Context, string) (*adt.RollbackResult, error) {
+	panic("not implemented")
+}
 func (m *mockClient) ReleaseTransportWithTasks(context.Context, string) error {
 	panic("not implemented")
 }
--- a/adt/registry.go
+++ b/adt/registry.go
@@ -182,6 +182,9 @@ func (r *ClientRegistry) ReleaseTransportWithTasks(ctx context.Context, transpor
 func (r *ClientRegistry) ReleaseTransportVerified(ctx context.Context, transportNumber string, includeTasks bool) (*ReleaseResult, error) {
 	return r.activeClient().ReleaseTransportVerified(ctx, transportNumber, includeTasks)
 }
+func (r *ClientRegistry) RollbackTransport(ctx context.Context, transportNumber string) (*RollbackResult, error) {
+	return r.activeClient().RollbackTransport(ctx, transportNumber)
+}
 func (r *ClientRegistry) GetTransportInfo(ctx context.Context, transportNumber string) (*TransportRequest, error) {
 	return r.activeClient().GetTransportInfo(ctx, transportNumber)
 }
--- a/adt/rollback.go
+++ b/adt/rollback.go
@@ -1,0 +1,149 @@
+package adt
+
+import (
+	"context"
+	"fmt"
+	"strings"
+)
+
+// RollbackEntry is one object processed by RollbackTransport.
+type RollbackEntry struct {
+	Type   string `json:"type"`
+	Name   string `json:"name"`
+	Reason string `json:"reason,omitempty"`
+}
+
+// RollbackResult reports the outcome of RollbackTransport, partitioning the
+// transport's objects into those restored, skipped, and failed.
+type RollbackResult struct {
+	Restored []RollbackEntry `json:"restored"`
+	Skipped  []RollbackEntry `json:"skipped"`
+	Failed   []RollbackEntry `json:"failed"`
+}
+
+// rollbackSourceTypes is the set of object types RollbackTransport restores
+// source for. DDIC types (TABL, DTEL, ...) are intentionally excluded and
+// skipped — they have no source to roll back the same way.
+var rollbackSourceTypes = map[string]bool{
+	"PROG": true,
+	"CLAS": true,
+	"INTF": true,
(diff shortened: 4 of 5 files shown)


hf-kklein:
## GO with notes

Reviewed independently against the consumer's current `tools/rollback.go` (doRollback/rollbackObject), adtler #85, and aibap.mcp #416. The port is faithful and the change is well-scoped. Local results below.

### Build & test (PATH = goroot 1.25 + go/bin)
- `go test ./...` → all packages **ok** (adt, adtxml, custexport, auth)
- `go test -run 'Rollback|FindPreTransport' -v ./adt/` → all subtests **PASS**
- `go vet ./...` → clean
- `gofmt -l adt/` → clean
- `golangci-lint run --enable dupl,goconst,gocyclo ./...` → **0 issues**
- `go build -tags integration ./adt/...` → compiles

### Port fidelity (line-by-line vs consumer)
Exact equivalence confirmed on every point that matters:
- `PgmID != "R3TR"` → skip `"not R3TR"` ✔
- `rollbackSourceTypes` = {PROG,CLAS,INTF,FUGR}; DDIC skipped `"non-source object type"` ✔
- `ObjectURI` error → skip-with-reason ✔
- version walk via `seenTransport`, two distinct errors (transport-not-seen vs no-prior-version) — identical strings ✔
- `GetVersionSource(restoreURI)` → `LockObject` → `defer UnlockObject` → `GetSource` for etag → `SetSource(ctx, uri, oldSource, lockHandle, "", current.ETag)` (empty transport arg preserved; matches `SetSource(objectURI, source, lockHandle, transport, etag)` at source.go:331) ✔
- `ActivateObjects([]{uri})`, `!Success` → join `m.Text` with `"; "` ✔
- error-wrapping prefixes all match the consumer verbatim ✔

`findPreTransportVersion` is a faithful extraction of the inline loop: consumer's `break`-on-first-match equals the helper's `return`-on-first-`seenTransport`; both error branches preserved.

### One subtle, low-risk divergence (non-blocking)
If the pre-transport version entry has an **empty** `ContentURI`, behavior differs: the consumer's `restoreURI == ""` guard reports `"no version before transport ..."`, whereas `findPreTransportVersion` returns `""` as success and the caller then fails later in `GetVersionSource("")` with a different message. Both still fail (no silent wrong-restore), only the surfaced error text differs, and an empty ContentURI on a real version feed is not expected. Worth a one-line guard or comment if you want byte-for-byte parity, but not a merge blocker.

### Version-history ordering assumption
The heuristic depends on `GetVersionHistory` returning newest-first. `parseVersionFeed` (version.go) emits entries in document order from the ADT atom feed (no re-sort), which SAP delivers newest-first — the same assumption the consumer already relies on in production. The newest-first contract is now documented in the `findPreTransportVersion` doc comment; consider also noting it on `GetVersionHistory` itself since the correctness of this feature hinges on it. **Caveat for CLAS**: `getClassVersionHistory` concatenates definitions+implementations feeds (`all = append(all, ...)`), so for classes the combined slice is newest-first *per include* but not globally interleaved. The transport-match walk still works as long as the target transport's entries and the immediately-prior entry fall within the same include's run — true in the common case, but classes are the most fragile path here and are exactly what the live #416 reproducer should exercise.

### Test quality
- `TestFindPreTransportVersion` covers the four meaningful cases (after-transport, multi-version span, transport-absent, created-by-transport). Good.
- `TestRollbackTransport_SkipsNonRestorable` httptest fixture is faithful to `parseTransportObjectsXML`: request-level `tm:abap_object` with `pgmid/type/name` attrs parse correctly (Go xml matches attr local names ignoring the `tm:` prefix). Asserts both skip reasons and zero restore work.
- The full lock→set→activate restore path is untested at unit level — acceptably deferred to the #416 consumer reproducer per the PR description, since it only composes already-integration-tested primitives. The CLAS interleaving caveat above is the one thing I'd want that live reproducer to hit.

### Interface / mock impact
- `RollbackTransport` added to `TransportClient`; implementers all compile: `httpClient` (rollback.go), `ClientRegistry` (registry.go:185), custexport mock (export_test.go) — confirmed by green build.
- **Consumer mock needing the method in the #416 bump**: `tools/source_test.go:30` — `type mockClient struct` is the only full `adt.Client` double in aibap.mcp (verified by grep; transport methods cluster at source_test.go:255–271, alongside the existing `ReleaseTransportVerified`). Add `func (m *mockClient) RollbackTransport(context.Context, string) (*adt.RollbackResult, error)` there or the bump won't compile.

### Verdict
Satisfies #85 and unblocks #416. Faithful port, clean idiom, lint/vet/test/gofmt all green. Approving as **GO with notes** — the empty-ContentURI divergence and the CLAS multi-include ordering caveat are worth a glance (and a reproducer pass on a class) but neither blocks merge.

PR #88, opened 2026-06-18
feat: VerifySource — syntax-check standalone ABAP source via a throwaway $TMP program

### LINKED ISSUE #87: feat: VerifySource — syntax-check standalone ABAP source via a throwaway $TMP program
## Summary

Add `VerifySource(ctx, source string) (valid bool, messages []SyntaxMessage, err error)` to `QualityClient`. It creates a temporary program in `$TMP`, writes the source, runs a syntax check against the inactive version, and deletes the temp program. `valid` is true when there are no error-severity (`"E"`) messages.

## Motivation

mcp-server-abap's `verify_source` tool implements this whole create→lock→set→check→delete dance in the MCP layer (~8 HTTP round-trips), including the hardcoded `$TMP` package, the temp-program naming, and the `Type=="E"` severity mapping. That is an ADT idiom and belongs here.

The inline-checkrun alternative (checking source passed in the request body) was investigated in mcp-server-abap#126 and found unavailable on both ECC and S/4 (the endpoint ignores the inline body / rejects `_action=CHECK`), so the throwaway-`$TMP` round-trip is the portable implementation.

Consumer issue: Hochfrequenz/aibap.mcp#415

hf-kklein:
## What

Adds `VerifySource(ctx, source string) (valid bool, messages []SyntaxMessage, err error)` (`adt/verify.go`), on `QualityClient`. It creates a temporary program in `$TMP`, writes the source, runs an inactive-version syntax check, and deletes the temp program (defer). `valid` is true when there are no `"E"` messages. Added to `ClientRegistry` and the custexport mock.

## Why

mcp-server-abap's `verify_source` tool implements this ~8-call create→lock→set-source→check→delete dance inline, with the hardcoded `$TMP` package, temp-program naming, and `Type=="E"` mapping — an ADT idiom that belongs here. The inline-checkrun alternative was found unavailable on ECC/S4 (mcp-server-abap#126), so the throwaway-`$TMP` round-trip is the portable implementation.

Closes #87. Consumer: Hochfrequenz/aibap.mcp#415.

## Tests

`adt/verify_test.go` mocks the full round-trip (CSRF, create, lock, get-source-for-etag, set-source, unlock, /checkruns, delete). The `/checkruns` handler echoes the requested object URI as the report's `triggeringUri` (which `SyntaxCheck` correlates on — the program name is random) and conditionally includes one `E` message. Covers `valid=true` (no messages) and `valid=false` (one `E`).

No live integration test here: the dance composes already-integration-tested primitives; the end-to-end behavior is exercised by the #415 consumer reproducer on bump (valid + deliberately-broken source on hfq/s4u).

## Notes

- Temp name `Z_ADTLER_VERIFY_%06d` (random; not security-sensitive — `//nolint:gosec`).
- Interface addition to `QualityClient`: the consumer's `mockClient` will need `VerifySource` in the #415 bump.

### DIFF
--- a/adt/client.go
+++ b/adt/client.go
@@ -95,6 +95,7 @@ type DumpClient interface {
 type QualityClient interface {
 	SyntaxCheck(ctx context.Context, objectURI string) ([]SyntaxMessage, error)
 	BatchSyntaxCheck(ctx context.Context, objectURIs []string) []ObjectSyntaxResult
+	VerifySource(ctx context.Context, source string) (valid bool, messages []SyntaxMessage, err error)
 	RunUnitTests(ctx context.Context, objectURI string, timeoutSeconds int) (*TestResult, error)
 	RunATCCheck(ctx context.Context, objectURIs []string, checkVariant string) (*ATCResult, error)
 	GetATCCustomizing(ctx context.Context) (*ATCCustomizingResult, error)
--- a/adt/client_test.go
+++ b/adt/client_test.go
@@ -26,6 +26,9 @@ const logoffPath = "/sap/public/bc/icf/logoff"
 // used by SearchObjects/SearchPackages tests. Hoisted for goconst.
 const searchEndpoint = "/sap/bc/adt/repository/informationsystem/search"
 
+// programsEndpoint is the ADT programs collection endpoint. Hoisted for goconst.
+const programsEndpoint = "/sap/bc/adt/programs/programs"
+
 // progType is the ADT object type for ABAP programs, used in assertions
 // across multiple test files. Hoisted for goconst.
 const progType = "PROG/P"
--- a/adt/custexport/export_test.go
+++ b/adt/custexport/export_test.go
@@ -64,6 +64,9 @@ func (m *mockClient) GetObjectInfo(context.Context, string) (*adt.ObjectInfo, er
 func (m *mockClient) SyntaxCheck(context.Context, string) ([]adt.SyntaxMessage, error) {
 	panic("not implemented")
 }
+func (m *mockClient) VerifySource(context.Context, string) (bool, []adt.SyntaxMessage, error) {
+	panic("not implemented")
+}
 func (m *mockClient) BatchSyntaxCheck(context.Context, []string) []adt.ObjectSyntaxResult {
 	panic("not implemented")
 }
--- a/adt/registry.go
+++ b/adt/registry.go
@@ -134,6 +134,9 @@ func (r *ClientRegistry) GetObjectInfo(ctx context.Context, objectURI string) (*
 func (r *ClientRegistry) SyntaxCheck(ctx context.Context, objectURI string) ([]SyntaxMessage, error) {
 	return r.activeClient().SyntaxCheck(ctx, objectURI)
 }
+func (r *ClientRegistry) VerifySource(ctx context.Context, source string) (bool, []SyntaxMessage, error) {
+	return r.activeClient().VerifySource(ctx, source)
+}
 func (r *ClientRegistry) BatchSyntaxCheck(ctx context.Context, objectURIs []string) []ObjectSyntaxResult {
 	return r.activeClient().BatchSyntaxCheck(ctx, objectURIs)
 }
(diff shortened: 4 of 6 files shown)


hf-kklein:
## Independent Review: GO

Reviewed purely on the merits with no prior development context. Verdict: **GO** — merge-ready.

### Build & test (Go 1.x at $HOME/goroot, golangci-lint at $HOME/go/bin)
| Check | Result |
|---|---|
| `go test ./...` | ok (all 4 packages) |
| `go test -run TestVerifySource -v ./adt/` | PASS (both subtests) |
| `go vet ./...` | clean |
| `gofmt -l adt/` | no output |
| `go build -tags integration ./adt/...` | compiles |

(Branch was checked out in the main adtler dir, not a worktree — reviewed in place.)

### Port fidelity — faithful
`adt/verify.go` is a step-by-step match of the consumer's `tools/verify.go` `verifySource`:
- `Z_ADTLER_VERIFY_%06d` random temp name (was `Z_MCP_VERIFY` — cosmetic/internal, fine), `ObjectURI("PROG", name)`, `CreateObject("PROG", name, "$TMP", …, "")`.
- `defer` lock+delete cleanup, registered **after** CreateObject succeeds.
- Ordering: `LockObject` → `GetSource` (for ETag) → `SetSource(…, lockHandle, "", src.ETag)` → `UnlockObject`.
- Unlock-on-error in both the get-source and set-source failure paths.
- `SyntaxCheck` → `valid = no "E" message`.
- Equivalent error wrapping. One minor difference: `rand.Intn(1000000)` vs the consumer's `999999` — inconsequential.

### Resource-leak / correctness — sound
Walked each early return:
- **CreateObject fails** → returns before the defer is registered (defer is the next statement). No temp object exists to leak. Matches the consumer.
- **Lock fails** → defer (re-lock + delete) still cleans up; the failed lock returned no handle, nothing to unlock.
- **GetSource fails** → explicit `UnlockObject` before return, then defer deletes. No held lock, no leaked object.
- **SetSource fails** → same: explicit unlock, then defer deletes.
- **SyntaxCheck fails** → lock already released on the happy path before the check; defer deletes.
No path creates the temp program and fails to delete it.

### Test quality — meaningful, not papering over
- The `/checkruns` mock echoes the request's `adtcore:uri` back as `triggeringUri`. Verified against `batchSyntaxCheckChunkWithVersion` (syntaxcheck.go:145-149), which indexes reports by `TriggerURI` and correlates by exact match on the requested URI — random program name and all. Faithful, not a bug workaround.
- valid vs invalid is genuinely distinguished: the `withError` branch injects exactly one `E` message; assertions check `valid` **and** message count/type.
- **False-positive retry not accidentally triggered**: `isFalsePositiveReportMissing` (syntaxcheck.go:55-66) requires len==1, Type==E, **Line==1, Column==0**, and text containing REPORT/PROGRAM + missing/fehlt. The mock's E message is at `#start=2,5` with text "Field FOO is unknown" — fails the line/col and keyword checks, so the active-version retry path does not fire. The invalid test correctly exercises the inactive path only.

### gosec/lint
`//nolint:gosec` on the `math/rand` temp name is appropriate (non-cryptographic, throwaway object name) and lint is clean (0 issues).

### Interface / mock impact
`VerifySource` added to `QualityClient`; all three in-repo implementers compile: `httpClient` (the impl), `ClientRegistry` (delegates to `activeClient()`), and the custexport `mockClient` (panic stub). Consumer impact for the #415 bump: **`aibap.mcp/tools/source_test.go:127-144`** — `mockClient` is a test double for `adt.Client` (which embeds `QualityClient`), so it must gain a `VerifySource` method or that package won't compile. That's the only consumer mock to touch.

### Satisfies #87 / unblocks #415
Yes. The API signature matches #87 exactly, relocates the full $TMP orchestration per #415's 'orchestration relocation' strategy, and #126 confirms inline-checkrun is unavailable on ECC/S4 so the throwaway round-trip is the correct portable approach. No live integration test here is acceptable per the PR rationale — primitives are already integration-tested and the #415 consumer reproducer exercises end-to-end on bump.

### Notes (non-blocking)
- No live integration test in this PR; relies on the #415 reproducer at bump time. Consistent with the stated workflow.
- `rand.Intn(1000000)` collision probability is negligible for throwaway temp objects; no concern.

PR #92, opened 2026-06-18
test: integration tests for VerifySource, GetObjectDependencies, SearchPackages

### LINKED ISSUE #91: test: add + run integration tests for the v0.3.x extracted functions (VerifySource, GetObjectDependencies, SearchPackages, ...)
## Goal

v0.3.0–v0.3.7 added a batch of functions extracted from mcp-server-abap (the SAP/ADT domain logic that previously lived in the MCP wrapper). They all have **unit tests**, but most have **no integration test** against a live SAP system. Since adtler owns the ADT HTTP/XML layer, this is the right place to validate their real behavior on R/3 + S/4.

This issue tracks adding (and running) `eachSystem`-based integration tests for the new functions.

## Background — what shipped, and current coverage

| Function (version) | adtler issue | Integration test today |
|---|---|---|
| `ObjectURI` (v0.3.0) | #75 | n/a — pure function, no HTTP |
| `SearchPackages` + `ObjectTypePackage` (v0.3.1) | #77 | ❌ none (thin wrapper over `SearchObjects`) |
| `TextElementLockURI` (v0.3.2) | #79 | n/a — pure function, no HTTP |
| `ClassifyError` + `ExceptionType*` (v0.3.3) | #81 | n/a — pure classification of an `*ADTError` |
| `ReleaseTransportVerified` (v0.3.4) | #83 | ❌ none (destructive) |
| `RollbackTransport` (v0.3.5) | #85 | ❌ none (destructive) |
| `VerifySource` (v0.3.6) | #87 | ❌ none (safe: creates+deletes `$TMP`) |
| `GetObjectDependencies` / `DependencyClient` (v0.3.7) | #89 | ❌ none (read-only) |

The three pure functions need no integration test. The remaining five do.

## Proposed integration tests (use `eachSystem(t)`, `//go:build integration`, `Z_ADT_MCP_TEST` fixtures)

**Safe / non-destructive — add these:**
- `VerifySource` — one call with valid ABAP (`valid == true`, no messages) and one with deliberately-broken ABAP (`valid == false`, has an `E` message). Self-cleaning (`$TMP`). Validates the create→set→check→delete round-trip on both systems.
- `GetObjectDependencies` — read-only; mirror the coverage the consumer had:
  - `PROG` against `Z_ADT_MCP_TEST_REPORT` (D010TAB flat lookup + DD02L/TADIR classification).
  - `TABL` against a standard table with known deps (e.g. `SCARR`) — exercises the DDIC BFS (DD03L→DTEL/CHECKTABLE).
  - Optionally `FUGR`/`FUNC`/`CLAS` to cover pool-program naming + SEOMETAREL.
- `SearchPackages` — search `Z*` (or `Z_ADT_MCP_TEST*`) and assert results are package (`DEVC/K`) objects. Thin, but confirms the `objectType` constraint reaches the wire on both systems.

**Destructive — guarded or manual:**
- `ReleaseTransportVerified` — the value is confirming the ECC "200-but-still-modifiable" silent-fail path (`Released == false`). Needs a disposable transport. Either a guarded test that creates its own throwaway transport and only runs when an explicit env flag is set, or a documented manual reproducer. Do NOT run against shared transports.
- `RollbackTransport` — needs a transport that touched a test object with version history. Same caution; guarded or manual.

## Kickstart prompt (paste into a fresh Claude Code session in the adtler repo)

> adtler v0.3.0–v0.3.7 added functions extracted from mcp-server-abap. They have unit tests but most lack integration tests against live SAP. I want to add and run `eachSystem`-based integration tests for the new behavior, on `hfq` (R/3) and `s4u` (S/4).
>
> Prereqs to confirm first: on `main`, VPN up, `~/.config/sap-mcp/systems.json` lists `hfq` and `s4u`, and the `Z_ADT_MCP_TEST` package is installed on both. Run `go test -tags=integration ./adt/...` once to confirm the existing suite is green/skips cleanly.
>
> Then add new `//go:build integration` tests (follow the existing `*_integration_test.go` style and `eachSystem(t)` helper) for the **non-destructive** new functions, and run them with `SAP_INTEGRATION_SYSTEMS=hfq,s4u go test -tags=integration -run '<TestName>' -v ./adt/...`:
> 1. `VerifySource` — valid ABAP → `valid==true`, no messages; broken ABAP → `valid==false` with an `E` message. (Self-cleaning `$TMP`.)
> 2. `GetObjectDependencies` — PROG on `Z_ADT_MCP_TEST_REPORT`; TABL on `SCARR` (asserts the DDIC BFS returns DTEL/table deps with sensible `use_type`s). Add FUGR/FUNC/CLAS if a suitable fixture exists. Assert the result shape (`object_type`/`object_name`/`count`/`dependencies[]{name,use_type}`/`warnings`).
> 3. `SearchPackages` — `Z*` returns objects whose type is `DEVC/K`.
>
> For the **destructive** `ReleaseTransportVerified` and `RollbackTransport`: do NOT run them against real transports automatically. Either write a guarded test that creates its own disposable transport + object and is opt-in via an env flag, or just document a manual reproducer and skip. Ask me before doing anything that releases or rolls back a real transport.
>
> Report pass/skip/fail per test per system. If a test fails, capture the request/response and report it before proposing any production-code change. The whole point is to confirm the v0.3.x extraction behaves identically to what mcp-server-abap did before — flag any divergence (e.g. the `VerifySource` temp-program URI is lower-cased; `GetObjectDependencies` clamps `maxDepth` to [1,10] with `<1→1`).

## Checklist

Add + run (non-destructive):
- [ ] `VerifySource` integration test (valid + invalid), green on hfq and s4u.
- [ ] `GetObjectDependencies` integration test — PROG (`Z_ADT_MCP_TEST_REPORT`) + TABL (`SCARR`); optionally FUGR/FUNC/CLAS.
- [ ] `SearchPackages` integration test (`Z*` → `DEVC/K`).
- [ ] Existing suite still green: `SAP_INTEGRATION_SYSTEMS=hfq,s4u go test -tags=integration ./adt/...`.

Destructive (guarded/manual, optional):
- [ ] `ReleaseTransportVerified` — silent-fail (`Released==false`) on ECC via a disposable transport, or documented manual reproducer.
- [ ] `RollbackTransport` — restore-to-pre-transport via a disposable transport, or documented manual reproducer.

Sign-off:
- [ ] No behavior divergence from the pre-extraction mcp-server-abap implementation.
- [ ] Results summarized in a comment here.

---

_Consumer-side counterpart: Hochfrequenz/aibap.mcp#426 (verifies the thin MCP wrappers). The pure functions `ObjectURI`/`TextElementLockURI`/`ClassifyError` are intentionally excluded — they have no HTTP path and are fully unit-tested. I'm happy to write these integration tests as a PR if that's preferred over an ad-hoc session._

hf-kklein:
## What

Adds `//go:build integration`, `eachSystem(t)`-based integration tests for the **non-destructive** functions extracted in v0.3.1/v0.3.6/v0.3.7, which previously had only unit tests:

- **`VerifySource`** (`verify_integration_test.go`) — valid source → `valid=true`; broken source (unknown type) → `valid=false` with an `E` message. Self-cleaning `$TMP`; no fixture or transport needed.
- **`GetObjectDependencies`** (`dependencies_integration_test.go`) — `PROG` on the `Z_ADT_MCP_TEST_REPORT` fixture (asserts the D010TAB lookup + classification run cleanly and `Count == len(Dependencies)`), and `TABL` on standard `SCARR` (asserts the DDIC BFS returns ≥1 `DATA_ELEMENT`).
- **`SearchPackages`** (`search_packages_integration_test.go`) — `Z*` returns only `DEVC`-typed objects and includes the `Z_ADT_MCP_TEST` fixture package.

The destructive `ReleaseTransportVerified` / `RollbackTransport` are intentionally **not** included (guarded/manual per #91).

Closes #91. Consumer counterpart: Hochfrequenz/aibap.mcp#426.

## How to run

```bash
SAP_INTEGRATION_SYSTEMS=hfq,s4u go test -tags=integration \
  -run 'VerifySource_Integration|GetObjectDependencies_Integration|SearchPackages_Integration' \
  -v ./adt/...
```

## Reviewer note — CI does NOT exercise these

The lint job runs without `--build-tags` and no CI job runs `-tags=integration`, so **CI neither compiles nor lints these files**. I verified locally:
- `go build -tags integration ./adt/...` — compiles
- `go test -tags integration -run '^$' ./adt/...` — test binary compiles

Please re-verify the `-tags integration` compile in review (a broken integration test would pass CI silently). The bare `"Z_ADT_MCP_TEST_REPORT"` literal matches the existing convention in `debugger/search/transport_integration_test.go` (a repo-wide `testReportName` const would be a separate cleanup; goconst doesn't run on tagged files in CI).

## Assumptions to confirm on first live run (I can't run against SAP here)

- `VerifySource` valid case assumes the ADT syntax check does not hard-error when the `REPORT` statement name differs from the random temp program name (true in practice, else `verify_source` would be unusable). If a system flags it, we adjust the fixture source.
- `SCARR` exists on both target systems (it's used by the existing aibap.mcp `TestIntegration_GetObjectDependencies_TABL_SCARR`).
- `SearchPackages` result `Type` is prefixed `DEVC`.

If any assertion is wrong on real SAP, the failure output tells us whether it's a test-expectation tweak or a genuine regression — report before changing production code.


hf-kklein:
## Verdict: **GO with notes**

Independent review, no live SAP access — focus was the compile check (the thing CI silently skips), conventions, and assertion soundness.

### Compile / build — all green (local, `go1.25.0`)
```
go build -tags integration ./adt/...        # exit 0
go vet  -tags integration ./adt/...          # exit 0, clean
go test -tags integration -run '^$' ./adt/...   # test binary compiles, no tests run
go test ./...                                # full normal suite ok (tagged files excluded)
```
No undefined symbols, no unused imports under the tag. All three method signatures match `adt.Client` (`VerifySource(ctx, source) (bool, []SyntaxMessage, error)`, `GetObjectDependencies(ctx, type, name, maxResults, maxDepth) (*DependencyResult, error)`, `SearchPackages(ctx, query, maxResults) ([]ObjectInfo, error)`). `adt.UseTypeDataElement` and `testPackage` resolve. This confirms the reviewer-note concern — the files compile cleanly under the tag.

### Conventions — correct
- `//go:build integration` + `package adt_test` on all three. ✓
- `eachSystem(t)` with the `sys := sys` capture and per-system `t.Run` sub-tests, matching the helper's documented usage. ✓
- Uses the shared `testPackage` const; the bare `"Z_ADT_MCP_TEST_REPORT"` literal matches existing convention (debugger/search/completions tests). ✓
- Only the 3 new files added; destructive `ReleaseTransportVerified`/`RollbackTransport` correctly excluded per #91; PR honestly flags that CI won't run these and lists the assumptions to confirm live. ✓

### Assertion soundness
- **VerifySource (valid)** — sound. `SetSource` writes non-empty valid source *before* `SyntaxCheck`, so an inactive version exists and the false-positive 'REPORT statement missing' retry in `SyntaxCheck` cannot fire. The REPORT-name vs random-temp-name mismatch does not hard-error in the ADT check. `valid==true` follows from no `E` messages.
- **VerifySource (broken)** — robust; unknown type reliably produces an `E` message on both systems.
- **GetObjectDependencies / PROG** — appropriately lenient (no error + header match + `Count==len(Dependencies)`, tolerates 0 deps). ✓
- **GetObjectDependencies / TABL SCARR** — `Count>0` + ≥1 `DATA_ELEMENT` is reasonable: SCARR fields carry `ROLLNAME` → DTEL via the DD03L BFS. SCARR is a safe standard-table assumption on R/3 and S/4 (already used by aibap.mcp's TABL test).

### Note for the live run (potential flake — not a blocker)
- **SearchPackages**: `Z*` with `maxResults=50` plus a hard assertion that `Z_ADT_MCP_TEST` is in the results. If a target system has **more than 50 `Z*` packages**, the fixture package may fall outside the returned window (results are server-ordered, typically alphabetical — and `Z_ADT_MCP_TEST` sorts late-ish), making `foundTestPackage` flake. On the first live run, if this fails, prefer narrowing the query to `Z_ADT_MCP_TEST*` (mirrors `TestSearchObjects_Wildcard`) or raising `maxResults`, rather than touching production code. The `strings.HasPrefix(r.Type, "DEVC")` check is correct (`ObjectTypePackage == "DEVC/K"`).

### Bottom line
No compile blockers — the highest-risk failure mode for tag-gated files is clear. Merge-ready pending the documented live confirmations; watch the SearchPackages window on the first run.

hf-mrdachner:
## Independent review — **NO-GO (close as superseded by #94)**

Reviewed with no prior conversation context, in a dedicated worktree, against `origin/main` as of today.

### Verification performed

| Check | Result |
|---|---|
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `go test ./...` | pass (all 4 packages) |
| `go build -tags integration ./adt/...` | pass |
| `go vet -tags integration ./adt/...` | pass |
| `golangci-lint run --build-tags integration --enable goconst ./adt/...` | **3 new goconst issues** (see below; `main` is clean) |
| Rebase onto current `main` | applies cleanly, still builds/vets/tests green |

So the branch being `BEHIND` is not a technical blocker — it rebases without conflict and compiles against today's `main`. The problem is not staleness, it is duplication.

### The core finding: #94 subsumes this PR

Both PRs close #91. For the three functions this PR covers, #94 has an equivalent or stronger test in every case, **and #94's tests have actually been executed against R/3 and S/4** (results table in its body). This PR states outright that it was never run against SAP and lists three unverified assumptions. Per CLAUDE.md ("a regression test is not one until you have seen it fail" — the same argument applies to a brand-new assertion nobody has watched go green or red), an unexecuted integration test is the weaker artifact of the two.

Beyond the overlap, #94 also carries three real `rollback.go` S/4 fixes (VRSD task numbers, `corrNr` on `SetSource`, unlock-before-activate) plus the destructive `ReleaseTransportVerified` / `RollbackTransport` tests behind the `transport` build tag — i.e. the remaining, unticked half of #91's checklist. Nothing here is a prerequisite for that work; the two sets of files simply collide.

### Per-function comparison

**`VerifySource`** — `adt/verify_integration_test.go:22-25`, `:35-36`
The valid fixture is a four-statement report whose `REPORT` name deliberately differs from the generated `$TMP` name, and the PR body admits this is an assumption ("If a system flags it, we adjust the fixture source"). #94 uses `REPORT zdummy.` — minimal, nothing to trip over — and its broken case is `IF` without `ENDIF`, chosen with an explicit note that a type conflict can come back `W` rather than `E` on some releases. The unknown-type approach here (`z_adtler_no_such_type_xyz`) is fine in practice but is a customer-namespace name whose absence is assumed rather than guaranteed. #94 is the safer fixture, verified on both systems.

**`GetObjectDependencies`** — `adt/dependencies_integration_test.go:31-33`, `:44-52`
`:31-33` explicitly tolerates zero dependencies for the PROG fixture. #94 asserts non-empty with a reason that holds up: the fixture is created and activated by `TestMain`/`setupFixtures` and references `cl_abap_unit_assert`, so an empty `D010TAB` result means a broken query path, not a sparse fixture. A test that passes on an empty result set is close to not testing the PROG path at all.

**The one thing this PR does better:** the SCARR case asserts at least one `adt.UseTypeDataElement` dependency (`:50-57`), using the exported constant. #94 only checks membership in a locally duplicated map of `UseType` string literals. That is a genuinely stronger and more idiomatic semantic assertion and is worth porting into #94 — it is, as far as I can tell, the only unique value in this PR.

**`SearchPackages`** — `adt/search_packages_integration_test.go:21-39`
Two issues:
1. `:26` hard-fails (`t.Fatalf`) when the fixture-prefixed search returns nothing. `adt/fixtures_integration_test.go` falls back to `$TMP` when the fixture package is absent (`fixturePackages`, and `setupFixtures` only *warns* on a failed `BrowsePackage`), so on a system without that package this test fails rather than skipping. CLAUDE.md asks integration tests to `t.Skip` when the fixture isn't there; #94 does exactly that.
2. `:31` uses `strings.HasPrefix(r.Type, "DEVC")`, which would also accept a hypothetical `DEVCx`. #94 compares against the exported `adt.ObjectTypePackage` (`"DEVC/K"`) — that is the constraint the function under test actually sets, so it is the right assertion.

Also structural: `adt/search_integration_test.go` already exists on `main`; #94 appends there rather than adding a fourth search-related file.

### Lint (informational, not CI-blocking)

Under `--build-tags integration`, `main` is goconst-clean; this branch introduces three hits, because `"Z_ADT_MCP_TEST_REPORT"` reaches its third occurrence:

- `adt/dependencies_integration_test.go:25`
- `adt/search_integration_test.go:26` (pre-existing, newly flagged)
- `adt/transport_integration_test.go:31` (pre-existing, newly flagged)

CI lints without build tags so this stays green, but `adt/integration_helpers_test.go:27` already defines `testReportURI` and is the obvious home for a `testReportName` const. #94 would trip the same rule, so whoever lands the survivor should hoist it.

Minor: `:34-36` and `:50-56` log every dependency's object name. CLAUDE.md asks integration tests to log counts and types, not object names, so CI logs stay clean. #94 caps its listing at 15 entries; a count plus a `UseType` histogram would be better still.

### Public-repository policy

The **diff is clean** — no hostnames, no `<SID>K9…` transport numbers, no registered-namespace object names, no client numbers. Only the project's own public `Z_ADT_MCP_TEST*` fixture names appear, which are provisioned by the suite itself.

The **PR body is not**: the "How to run" block spells out two `systems.json` aliases literally, and CLAUDE.md permits that only for the documented/defined `SAP_INTEGRATION_SYSTEMS` default set, not for prose or ad-hoc instructions. #94 writes `<r3-key>` / `<s4-key>` and is the model to follow. Issue #91 carries the same two aliases in several places; worth redacting there in passing, independently of what happens to this PR.

### Recommendation

1. **Close #92 as superseded by #94** — it adds no coverage #94 lacks, has not been run against a live system, and its `SearchPackages` and PROG assertions are the weaker of the two.
2. Before closing, port the SCARR `adt.UseTypeDataElement` assertion into #94's `TestGetObjectDependencies_TABL_Integration`, and consider swapping #94's `UseType` string-literal map for the exported `adt.UseType*` constants while there.
3. Redact the two system aliases from this PR's body and from issue #91.
4. #94 currently reports **no CI checks at all** on its branch — that needs sorting out before it merges, since it is the PR that touches production code (`adt/rollback.go`).

---

*Reviewed by an independent Claude Opus agent with no prior conversation context; verdict based solely on the diffs, the linked issue, CLAUDE.md, and the local build/test/lint runs above.*

PR #100, opened 2026-07-22
feat: classrun endpoint (RunClass) — spec, plan, and implementation

Jonatan-HF:
## Summary

Adds the `RunClass` client method for the ADT classrun endpoint
(`POST /sap/bc/adt/oo/classrun/{name}`), which executes any global class
implementing `IF_OO_ADT_CLASSRUN` and returns its console output — the generic
"Run as ABAP Application" primitive, headless. This PR now carries the **design
spec, the implementation plan, and the implementation** (originally opened as a
docs-only spec; implementation added on the same branch).

- Spec: `docs/superpowers/specs/2026-07-22-classrun-endpoint-design.md`
- Plan: `docs/superpowers/plans/2026-07-23-classrun-endpoint.md`

## Why (motivation)

This is the enabling primitive for headless recovery of **frozen enqueue
locks**, a real blocker for `aibap.mcp` (Hochfrequenz/aibap.mcp#383): a write
that fails with 403/423 can leave an enqueue lock held server-side that neither
`unlock_object` nor `force_unlock` can clear (connection-count 0 = orphaned;
only SM12's kernel module `enq_remove_locks` clears it). ADT has no native
endpoint to run the cleanup — `RunClass` is what lets a helper class (e.g.
calling `cl_enq_admin~remove_locks`) execute without a GUI.

## What's implemented

- `RunClass(ctx, className)` → `*ClassRunResult{ClassName, ConsoleOutput}`.
- `ClassRunClient` interface, embedded in the aggregate `Client` interface
  (and thus in `*ClientRegistry`; the `custexport` test double gained a stub).
- `Accept: text/plain` hard-coded; stateless POST with empty body; standard
  30 s `doMutate`; body decoded as `string(body)` so UTF-8 survives; namespace
  slash-encoding handled by the existing `encodeNamespacePath` path.
- Unit tests (httptest): success (POST, empty body, CSRF header, `text/plain`,
  parsed output), URI lower-casing, namespace `%2f` encoding, UTF-8 round-trip,
  HTTP error → `*ADTError`.
- Integration tests (`//go:build integration`, via `eachSystem(t)` over R/3 and
  S/4): real class run + throwing-class variant, with a `GetObjectInfo`
  existence pre-check for clean skips.

## Live handler verification (HFQ + S4U)

The endpoint contract was verified by reading the SAP handler
`CL_OO_ADT_RES_CLASSRUN` on both systems, not left to assumption:

- Endpoint present on **both** HFQ (ECC/R3) and S4U (resolves the spec's open
  verification point on ECC availability). The HFQ handler is an older variant
  (no `IF_OO_ADT_CLASSRUN_OUT`, uses `write_text`, also serves `PROG`/`DDLS`),
  but the `text/plain` client contract is identical.
- Response is `text/plain`, POST, no request body; the class name is uppercased
  server-side (so client lower-casing is a convention, not a requirement).
- **Failure signalling is split:** an uncaught runtime exception in `main()`
  propagates as a non-2xx → `ADTError`; "soft" failures (missing `S_DEVELOP`
  auth, interface not implemented, non-existent class) return **HTTP 200 with an
  error string in the body** — a missing class is 200-with-text, not 404.

## Test plan

- `go test ./...` — passing (4 new unit tests + existing suite).
- Integration tests are `//go:build integration` and require SAP credentials +
  the fixture classes (`ZCL_ADT_MCP_CLASSRUN_TST`, `ZCL_ADT_MCP_CLASSRUN_ERR`)
  delivered to `Z_ADT_MCP_TEST`. Pending the real-SAP run (see below).

## Pre-merge notes

- `needs:integration-test`: awaits the real-SAP run on R/3 + S/4 once the
  fixtures are delivered to `Z_ADT_MCP_TEST`.
- After that run pins the exact status code, `TestRunClass_ThrowingClass` will
  be tightened from a probe (logs the status) to a strict assertion.
- Adding `RunClass` to the exported `Client` interface is a minor-version API
  break for any external implementer of that interface — note in release notes.

## Companion

Consumed by the `aibap.mcp` `run_class` tool (Hochfrequenz/aibap.mcp#440),
which is `blocked-by-adtler` until this ships in a tagged release. Consumer-side
implications of the 200-with-text behaviour were cross-posted there.

### DIFF
--- a/adt/classrun_integration_test.go
+++ b/adt/classrun_integration_test.go
@@ -38,10 +38,13 @@ func TestRunClass_Integration(t *testing.T) {
 	for _, sys := range eachSystem(t) {
 		sys := sys
 		t.Run(sys.Name, func(t *testing.T) {
-			ctx := context.Background()
 			if _, err := sys.Client.GetObjectInfo(ctx, classrunClassURI(classrunFixture)); err != nil {
-				t.Skipf("classrun fixture %s not present on %s (deliver it to Z_ADT_MCP_TEST first): %v",
-					classrunFixture, sys.Name, err)
+				var adtErr *adt.ADTError
+				if errors.As(err, &adtErr) && adtErr.StatusCode == 404 {
+					t.Skipf("classrun fixture %s not present on %s (deliver it to Z_ADT_MCP_TEST first): %v",
+						classrunFixture, sys.Name, err)
+				}
+				t.Fatalf("GetObjectInfo pre-check for %s on %s failed: %v", classrunFixture, sys.Name, err)
 			}
 			result, err := sys.Client.RunClass(ctx, classrunFixture)
 			if err != nil {
--- a/adt/classrun_integration_test.go
+++ b/adt/classrun_integration_test.go
@@ -1,0 +1,99 @@
+//go:build integration
+
+package adt_test
+
+import (
+	"context"
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/Hochfrequenz/adtler/adt"
+)
+
+// classrunFixture is the global class in Z_ADT_MCP_TEST that implements
+// IF_OO_ADT_CLASSRUN and writes classrunFixtureOutput to the console.
+const (
+	classrunFixture       = "ZCL_ADT_MCP_CLASSRUN_TST"
+	classrunFixtureOutput = "CLASSRUN_OK"
+	classrunThrowFixture  = "ZCL_ADT_MCP_CLASSRUN_ERR"
+)
+
+// classrunClassURI returns the ADT class URI for a classrun fixture name.
+func classrunClassURI(name string) string {
+	return "/sap/bc/adt/oo/classes/" + strings.ToLower(name)
+}
+
+// TestRunClass_Integration runs a real classrun class on every whitelisted
+// system (R/3 and S/4 via eachSystem) and asserts the known console string
+// comes back. This also exercises the classrun framework on each system —
+// the endpoint handler CL_OO_ADT_RES_CLASSRUN is present on both HFQ/ECC and
+// S4U (spec open verification point #2, resolved).
+//
+// The fixture-existence pre-check uses GetObjectInfo, NOT a 404 from RunClass:
+// the handler returns HTTP 200 with an error string for a missing/invalid
+// class (verified against CL_OO_ADT_RES_CLASSRUN), so a missing fixture would
+// otherwise fail the ConsoleOutput assertion instead of skipping cleanly.
+func TestRunClass_Integration(t *testing.T) {
+	for _, sys := range eachSystem(t) {
+		sys := sys
+		t.Run(sys.Name, func(t *testing.T) {
+			ctx := context.Background()
+			if _, err := sys.Client.GetObjectInfo(ctx, classrunClassURI(classrunFixture)); err != nil {
+				t.Skipf("classrun fixture %s not present on %s (deliver it to Z_ADT_MCP_TEST first): %v",
+					classrunFixture, sys.Name, err)
+			}
+			result, err := sys.Client.RunClass(ctx, classrunFixture)
+			if err != nil {
+				t.Fatalf("RunClass on %s failed: %v", sys.Name, err)
+			}
+			if !strings.Contains(result.ConsoleOutput, classrunFixtureOutput) {
+				t.Errorf("%s: console output %q does not contain %q",
+					sys.Name, result.ConsoleOutput, classrunFixtureOutput)
+			}
+			t.Logf("%s classrun output: %q", sys.Name, result.ConsoleOutput)
+		})
+	}
+}
+
+// TestRunClass_ThrowingClass confirms how an uncaught runtime exception in
+// main() is signalled — spec open verification point #1. Verified against
+// CL_OO_ADT_RES_CLASSRUN: main() runs inside a TRY that catches only
+// cx_sy_create_object_error, so a genuine uncaught runtime exception (e.g.
+// cx_sy_zerodivide) propagates and the ADT REST framework turns it into a
+// non-2xx HTTP error -> *adt.ADTError. "Soft" failures (missing S_DEVELOP
+// auth, class does not implement the interface) instead come back as HTTP 200
+// with the error text in the body.
+//
+// The test therefore EXPECTS an ADTError for the throwing fixture, and logs
+// the observed status/body so the exact status code can be pinned. If the
+// fixture instead returns 200, that means it failed "softly" rather than
+// throwing — adjust the fixture, not the client.
+func TestRunClass_ThrowingClass(t *testing.T) {
+	for _, sys := range eachSystem(t) {
+		sys := sys
+		t.Run(sys.Name, func(t *testing.T) {
+			ctx := context.Background()
+			if _, err := sys.Client.GetObjectInfo(ctx, classrunClassURI(classrunThrowFixture)); err != nil {
+				t.Skipf("throwing fixture %s not present on %s: %v",
+					classrunThrowFixture, sys.Name, err)
+			}
+			result, err := sys.Client.RunClass(ctx, classrunThrowFixture)
+			if err == nil {
+				t.Logf("%s: throwing class returned HTTP 200 (soft failure), body: %q",
+					sys.Name, result.ConsoleOutput)
+				return
+			}
+			var adtErr *adt.ADTError
+			if !errors.As(err, &adtErr) {
+				t.Fatalf("%s: expected *adt.ADTError, got %T: %v", sys.Name, err, err)
+			}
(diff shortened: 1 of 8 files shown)


Jonatan-HF:
## Companion PRs & combined summary

This is a two-repo, spec-only change. Two Draft PRs, producer → consumer:

| PR | Repo | Role |
|----|------|------|
| **#100** (this) | adtler | `RunClass` classrun endpoint client (**producer**) |
| **Hochfrequenz/aibap.mcp#440** | aibap.mcp | `run_class` MCP tool (**consumer**, `blocked-by-adtler`) |

### Why — frozen enqueue locks are a real blocker for aibap.mcp

A write that fails with 403/423 can leave an enqueue lock held server-side that
**neither `unlock_object` nor `force_unlock` can clear** (connection-count 0 =
orphaned; only SM12's kernel module `enq_remove_locks` clears it). Reproduced
live on S4U: three failed RAP-BDEF writes left three orphaned `BDEF…` locks that
survived `force_unlock` — see Hochfrequenz/aibap.mcp#383. Every sequence editing
two coupled objects (RAP BDEF + behavior class, function group + module, …) can
hit this, and the only recovery today is manual SM12. ADT has no native endpoint
to run the cleanup. `RunClass`/`run_class` is the generic primitive that lets a
deployed helper (e.g. `cl_enq_admin~remove_locks`) run headlessly — the actual
lock-cleanup tool is a separate, later effort built on it.

### Rollout order

1. **This PR (adtler)** — review spec → implement `RunClass` (+ embed
   `ClassRunClient` in `Client`) → merge → tag.
2. **aibap.mcp#440** — bump adtler to that tag → add `run_class`
   (`blocked-by-adtler` until then).

hf-kklein (commented):
Independent review of the design (verified the technical claims against the current adtler code). **The endpoint design is sound and implementation-ready** — the notes below are mostly for the companion `run_class` consumer spec, not this PR.

### Verified against the codebase — claims hold
- `doMutate(ctx, method, path, body, headers)` auto-applies `encodeNamespacePath` (client.go:485 via `doMutateWith`), so the `//sap/bc/adt/oo/classrun//na2/foo` → `%2fna2%2ffoo` claim is correct — no manual escaping needed.
- The aggregate `Client` interface embeds every sub-client (`SourceClient`, `ObjectClient`, …), so embedding `ClassRunClient` there matches the established pattern exactly.
- `contentTypeTextPlain` exists and `doMutate` sets `X-CSRF-Token`, so the unit-test assertions are satisfiable as written.

The `RunClass` → `doMutate` → `checkResponse` → `string(body)` shape mirrors `GetSource`/`ActivateObjects` faithfully. No concerns with the endpoint client itself.

### For the consumer spec (`run_class` tool) — the real risk lives downstream, not here
Fresh live evidence from this week's #442/#383/#449 work is worth threading into the motivation, because it sharpens the scope:

1. **Lock-domain boundary.** `enq_remove_locks` operates on the runtime **ENQUEUE**. It will help with the #383/#449 orphaned-enqueue case, but it will **not** touch a **#442** "locked in request `<TR>`" conflict — that is a CTS object-directory registration, a different lock domain (cleared by retargeting the transport, now handled in aibap.mcp#450, or via SE09). Worth stating explicitly so nobody points `run_class`-based cleanup at a #442 lock and is surprised.

2. **End-to-end value is still unproven — and that's the thing to verify before declaring #383/#449 solved.** This session live-confirmed the *problem* (an orphaned DDLS edit-lock, `force_unlock` cleared 0, `lock_object`/`activate`/`delete` all 403'd — only SM12 could clear it; see aibap.mcp#449). What is **not** yet proven is that a `cl_enq_admin~remove_locks` helper invoked via `RunClass` actually clears that *specific* observed orphan class. That's the whole payoff of this initiative, so the `run_class` spec should carry an integration test that (a) manufactures an orphaned enqueue, (b) runs the helper, (c) asserts the lock is gone — not just that classrun executed. (The generic `RunClass` endpoint here is correctly agnostic to all of this.)

3. **Execution context / auth (confirm):** for the cleanup to reach a *prior* session's orphaned enqueue, classrun must run server-side under the authenticated user and `enq_remove_locks` must be able to remove that user's locks cross-session — which would be the key advantage over `force_unlock` (session-scoped). Worth confirming the helper's authorization needs (S_ENQUE-class) on both systems, since it's the difference between this working and 403-ing.

### Open verification points
Resolving them via the throwing/namespaced fixtures is the right call. FWIW points 1 (exception signalling) and 2 (`IF_OO_ADT_CLASSRUN` on ECC) are both resolvable on the live systems right now if you'd like them pinned before implementation rather than during — happy to run those probes.

Jonatan-HF:
### Integration test results — real SAP (R/3 + S/4)

Ran the `RunClass` integration tests against both HFQ (ECC/R3) and S4U (S/4HANA) after delivering the fixture classes to `$TMP` on each system.

#### Fixtures

| System | `ZCL_ADT_MCP_CLASSRUN_TST` | `ZCL_ADT_MCP_CLASSRUN_ERR` |
|---|---|---|
| HFQ (ECC/R3) | created (`write_text`), active | created (`RAISE EXCEPTION TYPE cx_sy_zerodivide`), active |
| S4U (S/4) | created (`write`), active | created but **left empty/inactive** — see note |

On S4U the `create_object` calls each left a session-bound edit enqueue (the known ESRDIRE case; `403 ExceptionResourceNoAccess`), which self-cleared only after the stateful session timed out (~11 min/object). `_TST` cleared in time; `_ERR` did not and remains empty/inactive.

#### `TestRunClass_Integration` — both systems PASS

```
--- PASS: TestRunClass_Integration/HFQ (0.40s)
    HFQ classrun output: "CLASSRUN_OK\n\n"
--- PASS: TestRunClass_Integration/S4U (0.39s)
    S4U classrun output: "CLASSRUN_OK\n\n"
```

`RunClass` executes a real classrun class and returns its console output verbatim on both R/3 and S/4. (Console output carries a trailing `\n\n` — no client-side trimming, as designed.)

#### `TestRunClass_ThrowingClass` — HFQ PASS, S4U not run

```
--- PASS: TestRunClass_ThrowingClass/HFQ (0.43s)
    HFQ: throwing class surfaced as ADTError, status 500: SAP returned an HTML error page (no parseable detail)
```

**Resolves open verification point #1:** an uncaught runtime exception (`cx_sy_zerodivide`) in `main()` surfaces as a non-2xx → `*adt.ADTError` with **HTTP 500**, not an HTTP-200-with-text soft failure. This matches the spec's hypothesis and the handler analysis (`CL_OO_ADT_RES_CLASSRUN` catches only `cx_sy_create_object_error`). The 500 body is an ABAP dump HTML page, so `parseADTError` reports "SAP returned an HTML error page (no parseable detail)" — the status code is the reliable signal here.

S4U's throwing case was skipped because its `_ERR` fixture never activated (ESRDIRE above); the pre-check `GetObjectInfo` correctly skipped rather than false-failed. The handler is identical in this respect on both systems, so ECC's result is representative, but S/4 confirmation is still outstanding.

#### Follow-ups (not blocking)

- Confirm the throwing case on S4U once `_ERR` is activated (its enqueue has since aged out).
- Tighten `TestRunClass_ThrowingClass` from a status-logging probe to a strict `status == 500` assertion now that the value is known (was deferred pending this run).
- Clean up the leftover empty/inactive `ZCL_ADT_MCP_CLASSRUN_ERR` on S4U.
- Permanent fixture delivery to the `Z_ADT_MCP_TEST` package/repo (these were `$TMP`, user-local, for this run).

Core behaviour is verified green on both systems. Removing `needs:integration-test`.

adt/classrun_integration_test.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"context"
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/Hochfrequenz/adtler/adt"
+)
+
+// classrunFixture is the global class in Z_ADT_MCP_TEST that implements
```
hf-kklein, line 14: Ist die im entsprechenden Package enthalten(dachte dass sei auch auf GitHub )

hf-kklein (approved):
Lgtm. Matthias ist der bessere reviewer

adt/classrun_integration_test.go
Jonatan-HF: Claude hate das für die Integrationstests in $TMP Pakete gepackt. Beim schreiben hat es erst auf S4U geschaut, da gibt es Z_ADT_MCP_TEST, beim anlegen der Testklassen dann aber im HFQ begonnen, da gibt es die nicht --> alles in $TMP

Jonatan-HF: Einfach weil ich diese dramatische Ausdrucksweise liebe :D

image

hf-kklein: Dafür bin ich bekannt

Issue #106, opened 2026-07-24
classrun (RunClass) runs the generated class load but never triggers generation → fresh classes fail with false 'does not implement main', re-activated classes run stale version


Jonatan-HF:
## Summary

`RunClass` (ADT classrun, `POST /sap/bc/adt/oo/classrun/{name}`) executes the class's **generated runtime load** and does not itself trigger load generation. Combined with the fact that the ADT activation endpoint (used by `ActivateObjects`) does not (re)generate that runtime load, this produces two user-visible defects when driving classrun purely over ADT REST (no Eclipse in the loop):

1. **Fresh class → false "does not implement" error.** A class created + activated purely via ADT REST that has never been generated-for-execution returns, from the classrun handler `CL_OO_ADT_RES_CLASSRUN`, **HTTP 200** with the body `Error: Class does not implement if_oo_adt_classrun~main method!` — even though the class *does* implement the interface (verified: SEOMETAREL RELTYPE=1, active, syntax-clean, activation succeeds). The handler's `CREATE OBJECT` raises `cx_sy_create_object_error` because the runtime load is not present; the handler masks this as the generic "does not implement main" soft-failure. No ST22 dump is produced.

2. **Changed class → stale execution.** After changing a class's source, re-activating over ADT REST (activation reports success, `GetSource` returns the new source), classrun keeps returning the **previously generated** version's output. The ADT re-activation does not invalidate/regenerate the runtime load.

Verified live on **S4U (SAP_BASIS 758)**, 2026-07-24.

## Not a DB/EML problem

An earlier bug report attributed defect 1 to database/RAP-EML access in the class. That is a **confound** — the real variable is runtime-load generation state, not what the class does. A trivial pure-`out->write` class with no DB access reproduces defect 1 identically, and an established (previously-generated) class with the same trivial body runs fine.

## Reproducer (S4U, SAP_BASIS 758)

Pure-console class, no DB access:

```abap
CLASS zcl_runclass_repro DEFINITION PUBLIC FINAL CREATE PUBLIC.
  PUBLIC SECTION.
    INTERFACES if_oo_adt_classrun.
ENDCLASS.
CLASS zcl_runclass_repro IMPLEMENTATION.
  METHOD if_oo_adt_classrun~main.
    out->write( 'V1' ).
  ENDMETHOD.
ENDCLASS.
```

1. Create + set source + activate the class purely via ADT REST (create object, PUT source, POST activation).
[...]

PR #115, opened 2026-08-05
fix(#114): RunClass runs on the long HTTP client with a default deadline

### LINKED ISSUE #114: RunClass is capped at 30s by the short HTTP client; mirror RunQuery (doMutateLong + default deadline)
## Summary

`RunClass` issues its POST through the 30-second HTTP client, even though executing an
`IF_OO_ADT_CLASSRUN` class is by definition open-ended ABAP. `RunQuery` already uses the
unlimited client with a sane default deadline. `RunClass` should do the same.

Consumers cannot work around this: `http.Client.Timeout` combines with the context deadline as
`min(Timeout, ctx)`, so a longer caller context cannot raise the cap, and the client fields are
unexported with no constructor parameter or setter.

Downstream: Hochfrequenz/aibap.mcp#465

## Root cause (v0.3.13)

```go
// adt/client.go:222
http: &http.Client{Timeout: 30 * time.Second, ...}
// adt/client.go:226
httpLong: &http.Client{Timeout: 0, ...}   // "no timeout; caller controls via context deadline"

// adt/classrun.go:45
resp, err := fresh.doMutate(ctx, http.MethodPost, uri, nil, ...)   // <- short client
```

Compare `RunQuery`, which gets it right (`adt/query.go:44-49`, `:57`):

```go
// Apply default timeout if caller didn't set one.
if _, hasDeadline := ctx.Deadline(); !hasDeadline {
    var cancel context.CancelFunc
    ctx, cancel = context.WithTimeout(ctx, 5*time.Minute)
    defer cancel()
}
...
resp, err := c.doMutateLong(ctx, "POST", path, strings.NewReader(trimmed), headers)
```

## Requested change

In `RunClass`, mirror `RunQuery` **in full** — both halves:

1. Apply a default context deadline when the caller has not set one.
2. Send via `doMutateLong` instead of `doMutate`.

Switching to `doMutateLong` alone would replace one hard limit with **no** limit, letting a
runaway classrun hang the calling session indefinitely. Both parts are required.

Open question for the implementer: whether the default should be `RunQuery`'s 5 minutes or
something shorter. `RunClass` runs arbitrary user-authored ABAP with side effects, so a
different default may be justified. Worth a deliberate decision rather than copying 5 minutes
by reflex.

Note `RunClass` calls `c.freshSession()` first — the fresh client must carry the same
`httpLong` configuration. `freshSession` already copies `httpLong` (`adt/client.go:263-266`),
so this should hold, but it needs a test rather than an assumption.

## Reproducer

Measured on S/4 (S4U), 2026-08-05.

Class in `$TMP`, active — busy loop with a hard cap so it terminates on its own:

```abap
CLASS zcl_aibap_timeout DEFINITION PUBLIC FINAL CREATE PUBLIC.
  PUBLIC SECTION.
    INTERFACES if_oo_adt_classrun.
  PRIVATE SECTION.
    CONSTANTS mc_cap_seconds TYPE i VALUE 45.
ENDCLASS.

CLASS zcl_aibap_timeout IMPLEMENTATION.
  METHOD if_oo_adt_classrun~main.
    DATA lv_counter TYPE int8.
    DATA lv_elapsed TYPE p LENGTH 8 DECIMALS 2.
    GET TIME STAMP FIELD DATA(lv_start).
    DO.
      lv_counter = lv_counter + 1.
      IF lv_counter MOD 5000000 = 0.
        GET TIME STAMP FIELD DATA(lv_now).
        lv_elapsed = cl_abap_tstmp=>subtract( tstmp1 = lv_now tstmp2 = lv_start ).
        IF lv_elapsed >= mc_cap_seconds.
          EXIT.
        ENDIF.
      ENDIF.
    ENDDO.
    out->write( |elapsed { lv_elapsed } s| ).
  ENDMETHOD.
ENDCLASS.
```

```go
res, err := client.RunClass(context.Background(), "ZCL_AIBAP_TIMEOUT")
```

**Current (broken):**

```
RunClass: Post ".../sap/bc/adt/oo/classrun/zcl_aibap_timeout":
context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

after ~30 s wall clock. The ABAP work process is still running at that point — this is the Go
client giving up, not SAP.

**Expected (fixed):**

```
res.ConsoleOutput == "elapsed 45.00 s\n"
```

Control measurement: the same class with `mc_cap_seconds = 20` returns cleanly today
(`elapsed 20.00 s`, 405,000,000 iterations), confirming the boundary is the client timeout and
not an ABAP runtime limit.

## Test requirements

1. **Unit** — `httptest` server that delays its response past 30 s (or an injectable delay);
   assert `RunClass` succeeds. Guards the actual regression.
2. **Unit** — caller-supplied deadline is honoured: `context.WithTimeout(ctx, 1*time.Second)`
   against a slow handler must fail with a context error. Guards against removing the limit
   entirely.
3. **Unit** — default deadline applies when the caller supplies none: a context without a
   deadline must not hang forever.
4. **Unit** — `freshSession()` preserves the long-client configuration, so `RunClass` does not
   silently fall back to the 30 s client via the fresh session path.
5. **Regression guard** — assert `RunQuery` and `RunClass` route through the same client, so
   the two cannot drift apart again.
6. **Integration** — the 45 s reproducer above against S/4, asserting clean console output
   rather than a timeout error.

Jonatan-HF:
Fixes #114. Consumer issue: Hochfrequenz/aibap.mcp#465

## What was wrong

`RunClass` issued its POST through the 30-second HTTP client, even though executing an
`IF_OO_ADT_CLASSRUN` class is arbitrary user-authored ABAP. Anything running longer than 30 s
failed with `Client.Timeout exceeded while awaiting headers` while the ABAP work process kept
going. Consumers could not work around it: `http.Client.Timeout` and the context deadline
combine as `min(Timeout, ctx)`, and the client fields are unexported.

## What changed

`RunClass` now mirrors `RunQuery` in full — both halves, as the issue requires:

1. the POST goes through `doMutateLong` (the client with no cap of its own), and
2. `withDefaultDeadline(ctx)` applies a default deadline when the caller supplies none.

Doing only (1) would trade a wrong limit for **no** limit and let a runaway classrun hang the
caller indefinitely.

The shared default is hoisted into `defaultLongRunTimeout` behind the new `withDefaultDeadline`
helper, which `RunQuery` now uses as well — one default for "open-ended ABAP", so the two
endpoints cannot drift apart again.

### The default: 5 minutes, deliberately

The issue asked for a decision rather than a reflex copy of `RunQuery`'s 5 minutes. 5 minutes
it is, on these grounds:

- classrun executes in a **dialog work process**, so SAP enforces `rdisp/max_wprun_time`
  (commonly 300-600 s) itself. A client deadline sitting just past that lets SAP abort the step
  and return a diagnosable error, which beats the client bailing out first and leaving the work
  process running — exactly the failure mode #114 is about.
- A shorter default (say 60 s) would re-create the reported bug for legitimately slow classruns,
  just at a different number.
- Callers who want tighter control pass a context deadline; theirs always wins.
- One shared value for both open-ended-ABAP endpoints is the drift guard the issue asked for.

## Tests

`adt/classrun_timeout_internal_test.go` is an internal (`package adt`) test on purpose: the
regression is about *which* `*http.Client` carries the request, and the only way to observe that
without sleeping 30+ s of real time is to shrink the short client's timeout to milliseconds. A
request that survives a 30 ms short-client cap cannot be on the short client.

| Test | Guards |
|---|---|
| `TestRunClass_NotCappedByShortClient` | the actual regression — verified to FAIL on the pre-fix code (`Client.Timeout exceeded`) |
| `TestRunClass_HonoursCallerDeadline` | the limit was not removed; caller deadline wins |
| `TestRunClass_DefaultDeadlineApplies` | a context without a deadline still gets one |
| `TestWithDefaultDeadline` | the helper itself, both branches |
| `TestFreshSession_PreservesLongClient` | `freshSession` keeps the long-client config (not assumed) and still gets its own cookie jar (#106 isolation) |
| `TestRunQueryAndRunClass_ShareLongPath` | drift guard: both route through the long client AND share one default — verified to FAIL on the pre-fix code |

Integration: `TestRunClass_LongRunning_Integration` creates the issue's 45-second busy-loop class
in `$TMP` on every whitelisted system, runs it with a plain `context.Background()` (so the
default deadline is what carries the run), and asserts both the console output and
`elapsed >= 30s` — a pass in under 30 s would mean the class never ran long enough to prove
anything. The probe class caps itself, so it terminates even if the client goes away, and the
fixture is deleted in cleanup.

## Verification

- `go test ./...` — pass
- `go build ./...`, `go vet ./...` — pass
- `go build -tags integration ./adt/...`, `go vet -tags integration ./adt/...` — pass
- Real-SAP integration test still outstanding (`needs:integration-test`)

Also documents the two-HTTP-client rule in CLAUDE.md so the next open-ended-ABAP endpoint gets
both halves from the start.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -104,6 +104,17 @@ R/3 (ECC) and S/4HANA often behave differently for the same ADT endpoint. Always
 
 The inverse also matters: some operations need a **fresh** session, not a reused one. `RunClass` runs on a single-use isolated session (`freshSession`) because a session that performed the create/set source/activate lifecycle cannot generate a class's runtime load on S/4 (see "Runtime-load generation vs. session reuse" under SAP system differences). If an operation depends on state SAP only generates in a clean session, give it a fresh session instead of reusing the caller's.
 
+### Long-running ABAP execution (two HTTP clients)
+
+The client holds two `*http.Client`s: `http` with a 30-second timeout for ordinary ADT calls, and `httpLong` with no timeout of its own, where the deadline comes from the context (`doReadLong`, `doMutateLong`).
+
+Any endpoint that executes **open-ended ABAP** — currently `RunQuery` (data preview) and `RunClass` (classrun) — must use both halves of the long path:
+
+1. `doMutateLong` / `doReadLong`, because 30 seconds is arbitrary for user-authored ABAP and consumers cannot raise it (`http.Client.Timeout` and the context deadline combine as `min(...)`, and the client fields are unexported), and
+2. `withDefaultDeadline(ctx)`, because the long client imposes no limit at all — without a default deadline a runaway statement hangs the caller forever.
+
+Doing only (1) trades a wrong limit for no limit. The shared default lives in `defaultLongRunTimeout` (5 minutes, just past the usual SAP dialog work-process limit so SAP aborts the step and returns a diagnosable error first); both endpoints reference it so they cannot drift apart. Fixed for `RunClass` in #114.
+
 ## Coding Pitfalls
 
 - **Never use Go backtick (raw) string literals for ABAP source code** in test fixtures. Backtick strings preserve tab indentation from the Go source file.
--- a/adt/classrun.go
+++ b/adt/classrun.go
@@ -21,6 +21,10 @@ type ClassRunClient interface {
 	// endpoint and returns whatever the class writes to the console handler
 	// (out->write(...)). It does not validate that the class exists, is
 	// active, or implements IF_OO_ADT_CLASSRUN — the caller pre-checks that.
+	//
+	// Executing a class is open-ended ABAP, so the run is not capped by the
+	// short HTTP client's 30-second timeout. Pass a context deadline to bound
+	// it; without one, defaultLongRunTimeout applies.
 	RunClass(ctx context.Context, className string) (*ClassRunResult, error)
 }
 
@@ -36,13 +40,24 @@ type ClassRunClient interface {
 // own concern — so an isolated session is consistent with its contract and
 // leaves the caller's session and locks untouched.
 //
+// The POST goes through the LONG-TIMEOUT HTTP client (doMutateLong): running an
+// IF_OO_ADT_CLASSRUN class is arbitrary user-authored ABAP and routinely takes
+// longer than the short client's 30-second cap, which consumers cannot raise
+// (http.Client.Timeout and the context deadline combine as min(...), issue
+// #114). Because that client imposes no limit of its own, a default deadline is
+// applied when the caller supplies none — otherwise a runaway class would hang
+// the caller indefinitely. Both halves are required; see defaultLongRunTimeout.
+//
 // Namespace slashes in className are percent-encoded automatically by
-// doMutate → encodeNamespacePath (triggered by the "//" that results from
+// doMutateLong → encodeNamespacePath (triggered by the "//" that results from
 // appending "/na2/foo" to the base).
 func (c *httpClient) RunClass(ctx context.Context, className string) (*ClassRunResult, error) {
+	ctx, cancel := withDefaultDeadline(ctx)
+	defer cancel()
+
 	fresh := c.freshSession()
 	uri := "/sap/bc/adt/oo/classrun/" + strings.ToLower(className)
-	resp, err := fresh.doMutate(ctx, http.MethodPost, uri, nil,
+	resp, err := fresh.doMutateLong(ctx, http.MethodPost, uri, nil,
 		map[string]string{"Accept": contentTypeTextPlain})
 	if err != nil {
 		return nil, fmt.Errorf("RunClass: %w", err)
--- a/adt/classrun_integration_test.go
+++ b/adt/classrun_integration_test.go
@@ -206,6 +206,112 @@ func TestRunClass_ReactivatedClass_Integration(t *testing.T) {
 	}
 }
 
+// longRunCapSeconds is how long the issue #114 probe class busy-loops before it
+// exits on its own — comfortably past the 30-second short-HTTP-client cap that
+// used to abort RunClass, and short enough to stay well inside the SAP dialog
+// work-process limit.
+const longRunCapSeconds = 45
+
+// longRunClassrunSource returns an IF_OO_ADT_CLASSRUN implementation that busy-
+// loops for capSeconds and then writes "elapsed <n> s". The hard cap means the
+// class terminates by itself even if the client goes away. Built from double-
+// quoted strings (not a Go backtick literal) per the CLAUDE.md ABAP-fixture rule.
+func longRunClassrunSource(className string, capSeconds int) string {
+	l := strings.ToLower(className)
+	return "CLASS " + l + " DEFINITION PUBLIC FINAL CREATE PUBLIC.\n" +
+		"  PUBLIC SECTION.\n" +
+		"    INTERFACES if_oo_adt_classrun.\n" +
+		"  PRIVATE SECTION.\n" +
+		"    CONSTANTS mc_cap_seconds TYPE i VALUE " + fmt.Sprint(capSeconds) + ".\n" +
+		"ENDCLASS.\n\n" +
+		"CLASS " + l + " IMPLEMENTATION.\n" +
+		"  METHOD if_oo_adt_classrun~main.\n" +
+		"    DATA lv_counter TYPE int8.\n" +
+		"    DATA lv_elapsed TYPE p LENGTH 8 DECIMALS 2.\n" +
+		"    GET TIME STAMP FIELD DATA(lv_start).\n" +
+		"    DO.\n" +
+		"      lv_counter = lv_counter + 1.\n" +
+		"      IF lv_counter MOD 5000000 = 0.\n" +
+		"        GET TIME STAMP FIELD DATA(lv_now).\n" +
+		"        lv_elapsed = cl_abap_tstmp=>subtract( tstmp1 = lv_now tstmp2 = lv_start ).\n" +
+		"        IF lv_elapsed >= mc_cap_seconds.\n" +
+		"          EXIT.\n" +
+		"        ENDIF.\n" +
+		"      ENDIF.\n" +
+		"    ENDDO.\n" +
+		"    out->write( |elapsed { lv_elapsed } s| ).\n" +
+		"  ENDMETHOD.\n" +
+		"ENDCLASS.\n"
+}
+
+// TestRunClass_LongRunning_Integration is the issue #114 regression test: a
+// classrun that runs for longRunCapSeconds (45 s) must come back with its
+// console output instead of dying at the 30-second short-HTTP-client cap.
+//
+// Pre-fix, RunClass POSTed through the 30 s client and failed with
+//
+//	Post ".../oo/classrun/...": context deadline exceeded
+//	(Client.Timeout exceeded while awaiting headers)
+//
+// after ~30 s wall clock, while the ABAP work process kept running. Post-fix the
+// POST goes through the long client with the default deadline, so the run
+// completes. The elapsed-time assertion is the actual proof: a pass in under
+// 30 s would mean the class did not really run long, making the test vacuous.
+//
+// The caller passes a plain context.Background() on purpose — the default
+// deadline must be what carries the run.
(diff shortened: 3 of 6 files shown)


Jonatan-HF:
## Review of #115 (fix/114-runclass-long-client, commit 5f091da)

Reviewed independently against issue #114 and the PR diff, with the working tree checked out at the target commit plus a scratch worktree used to verify pre-fix test failure.

### Verification performed

All commands run locally against the PR commit:

- `go build ./...` — pass
- `go test ./... -count=1` — pass (`adt` package ~13-15s, see note below on wall-clock cost)
- `go build -tags integration ./adt/...` — pass
- `go vet -tags integration ./adt/...` — pass

Not run: the real-SAP integration test (`TestRunClass_LongRunning_Integration`), consistent with the PR's own `needs:integration-test` label — no SAP credentials available in this review environment.

To confirm the new regression tests are not vacuous, I built a throwaway git worktree off `main` (pre-fix), copied in only `classrun_timeout_internal_test.go` plus the new `client.go`/`query.go` helpers while deliberately leaving `classrun.go`'s `RunClass` on the old `doMutate` path (i.e. "helper added, call site not fixed" — the exact bug the drift guard targets):

- `TestRunClass_NotCappedByShortClient` fails on pre-fix `RunClass`, with the exact pre-fix error text (`Client.Timeout exceeded while awaiting headers`).
- `TestRunQueryAndRunClass_ShareLongPath/both_bypass_the_short_client_cap` also fails in the same scenario.

Re-applying the real `classrun.go` change makes the whole new suite pass. So both regression tests are real, not tautological.

### Correctness

- `RunClass` now runs `withDefaultDeadline(ctx)` before `freshSession()` and posts via `fresh.doMutateLong`. `freshSession()` (client.go:254-272) correctly copies `httpLong.Timeout` (0) and `httpLong.Transport` from the parent, with its own fresh cookie jar shared between `http` and `httpLong` — verified by `TestFreshSession_PreservesLongClient`, and confirmed by reading the function directly.
- `doMutateWith`/`execMutateWith` (client.go:548-625) thread the passed-in `*http.Client` (`hc`) through both the initial attempt and the 401/403 CSRF-retry attempt, so the retry path does not silently fall back to the short client. This was a specific risk called out in the review brief and it holds up on inspection.
- `defer cancel()` in `RunClass` is safe: it's a plain synchronous function (no goroutine handoff), so `cancel()` fires only after the response body has been fully read and the function is returning — no premature-cancellation-before-body-read bug.
- `withDefaultDeadline` (client.go:517-529) is a small, correct helper: caller deadline wins (no-op cancel), otherwise applies `defaultLongRunTimeout`. `RunQuery` was refactored to use it too, removing its former ad-hoc copy of the same logic — a legitimate dedup, and the PR's "drift guard" framing is justified.
- The 5-minute default is reasonable and the PR's stated rationale (sit just past `rdisp/max_wprun_time` so SAP aborts the dialog step and returns a diagnosable error before the client just gives up) is sound and matches how R/3/S/4 dialog work processes behave. I don't see a better default to suggest; a shorter default would just recreate #114 for legitimately-slower classruns, and there's no endpoint-specific reason (e.g. classrun's side effects) that argues for shortening it below what `RunQuery` already used successfully in production.

### Test coverage against issue #114's 6 requirements

1. httptest delay past 30s equivalent — covered by `TestRunClass_NotCappedByShortClient` (shrinks the *short* client to 30ms instead of stretching real time to 30s+; verified above that it fails pre-fix).
2. Caller-supplied deadline honoured — `TestRunClass_HonoursCallerDeadline`, confirms `context.DeadlineExceeded` and a prompt (<1s) abort.
3. Default deadline applies when caller supplies none — `TestRunClass_DefaultDeadlineApplies`, shrinks the package var to 150ms and confirms it fires.
4. `freshSession()` preserves the long client — `TestFreshSession_PreservesLongClient`, checks `Timeout`, `Transport` identity, and jar isolation/sharing.
5. `RunQuery`/`RunClass` drift guard — `TestRunQueryAndRunClass_ShareLongPath`, two subtests (short-client bypass, shared default). Verified to fail pre-fix.
6. Integration — `TestRunClass_LongRunning_Integration`, a 45s busy-loop class in `$TMP`, run via `eachSystem(t)`, asserting both console output and `elapsed >= 30s` (so a fast-return pass can't slip through vacuously). Not executed in this review (no SAP access). The test creates its probe class via the existing `createTmpClassrunClass` helper, which already registers `t.Cleanup(... client.DeleteObject ...)` (classrun_integration_test.go:96-100) — so the fixture is deleted after the run and this new test gets that cleanup for free; no gap here.

`t.Parallel()` is not used anywhere in package `adt` (confirmed via grep), so mutating the package-level `defaultLongRunTimeout` var in tests, and restoring it via `defer`, carries no actual cross-test race risk under the current test suite — the risk the review brief asked about does not currently materialize, but it is a latent trap: if any future test in this package adopts `t.Parallel()`, these var-swapping tests would need `t.Parallel()`-safe handling (e.g. a setter that fails if called concurrently, or simply keep this file free of parallelism).

### Unit-test wall-clock / httptest-Close concern (non-blocking, but a real finding)

The review brief asked to check whether the sleeping httptest servers can leak or block in `Close`. They do block, longer than the test authors likely intended:

`TestRunQueryAndRunClass_ShareLongPath`'s second subtest ("both apply the same default deadline") spins up `slowServer(10 * time.Second)`, shortens `defaultLongRunTimeout` to 150ms, and expects the client side to bail out quickly. The client does bail out quickly (context deadline), but the server-side handler's `select { case <-time.After(delay): ... case <-r.Context().Done(): return }` does **not** unblock early in practice — running this test repeatedly reproduces Go's own diagnostic:

```
httptest.Server blocked in Close after 5 seconds, waiting for connections: ...in state active
```

and the subtest's actual wall time is ~10.1-10.6s, i.e. it waits out the full server-side sleep rather than returning in ~150ms as the comments imply. This is not an unbounded leak (it does eventually resolve when the handler's `time.After(delay)` fires), but it inflates the unit suite by roughly 10 seconds on every run of `go test ./...`, and the design intent (fast-cancelling handler via context) silently fails to do its job. If a future edit bumps that delay to something larger (e.g. testing a longer default), this becomes a much worse hidden cost or a CI flake/timeout risk. Recommend either accepting the ~10s cost explicitly (drop the now-nonfunctional `r.Context().Done()` case or note why it doesn't help), or restructuring the subtest so the server delay is proportional to the shortened deadline (e.g. a smaller delay, since the point under test is only "which client is used," not "does it actually take 10 wall-clock seconds").

### Style / conventions

- `longRunClassrunSource` in `classrun_integration_test.go` builds the ABAP fixture from `+`-concatenated double-quoted strings rather than a backtick literal, per the CLAUDE.md rule — correctly followed, and called out in its own doc comment.
- New constants (`discoveryPath`, `dataPreviewPath`, `tinyShortTimeout`, `slowResponse`, `emptyTableData`, `longRunCapSeconds`) are hoisted rather than inlined repeatedly, consistent with the `goconst` guidance in CLAUDE.md, and `golangci-lint` (with `goconst` enabled) confirms 0 issues.
- Doc comments on `RunClass`, `ClassRunClient.RunClass`, `defaultLongRunTimeout`, and `withDefaultDeadline` are clear, explain the "why" (not just the "what"), and correctly cross-reference issue #106 and #114.
- `query.go`'s unused `"time"` import was correctly dropped after delegating to `withDefaultDeadline`.
- The CLAUDE.md addition ("Long-running ABAP execution (two HTTP clients)") is accurate and matches the code as reviewed; it's a reasonable generalization for future open-ended-ABAP endpoints.

### Summary of findings

- Non-blocking: `TestRunQueryAndRunClass_ShareLongPath`'s second subtest doesn't actually cancel the server-side handler early despite the `r.Context().Done()` case, adding ~10s to every unit test run. Worth tightening but not correctness-affecting.
- No blocking issues found. Core fix (routing through `doMutateLong` + `withDefaultDeadline`, `freshSession` preserving `httpLong`, retry path preserving the client choice) is correct and well-tested; the regression tests were verified to actually fail pre-fix.

### Verdict

**GO with notes**

Jonatan-HF:
@C:/Users/JONATA~1/AppData/Local/Temp/claude/reply.md

Jonatan-HF:
@C:/Users/JONATA~1/AppData/Local/Temp/claude/integ.md

